PR #1, opened 2023-07-28
setup Prettier for code formatting

JafariM:


package.json
```diff
@@ -37,5 +37,15 @@
...
+  },
+  "devDependencies": {
+    "husky": "^8.0.3",
+    "prettier": "3.0.0",
+    "pretty-quick": "^3.1.3"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "pretty-quick --staged"
+    }
```
akosasante, line 51: cool, so this will use Prettier to format your changed files each time you commit? 🤩

package.json
akosasante, line 1: I believe that the `react-scripts` package also installs [ESLint ](https://eslint.org/)(you can check by running the command `npm list eslint` which will show any packages that are dependent on ESLint). ESLint checks for errors in your javascript code and can auto-fix them. But sometimes its auto-fixes conflict with Prettier's auto formatting. Because of that, it may be a good idea to also install [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier#installation) package. This video talks a bit about it in the second half: https://www.youtube.com/watch?v=DqfQ4DPnRqI

.prettierrc.json
akosasante, line 1: I tried to run Prettier on this branch using `npx prettier -c .` and it gave the following error:

```
Checking formatting...
[error] Invalid configuration for file "/workspaces/dd-prac-team4-front/.prettierrc.json":
[error] JSON Error in /workspaces/dd-prac-team4-front/.prettierrc.json:
[error] ' is not valid JSON while parsing '��
```

I think that's because it's expecting this file to be a JSON file with some rules in there. If you want to start off just using the default Prettier formatting rules, then you should make this file just contain an empty JSON object: `{}`

akosasante (commented):
Looking good! Added a few suggesions, and  requesting a fix for `.prettierrc.json`. It would be a good idea to update the `README.md` with a section on `Formatting` that just explains to any user of this repo that Prettier is installed and will auto-format files on git commits using the `Husky` tool.

.prettierrc.json
JafariM: I add an empty object to the .prettierrc.json file and run this command `npx prettier -c ` . It runs without any error in my end.

package.json
JafariM: Yes, it will

package.json
JafariM: Thank you, I just add that plugin to the repo

README.md
```diff
@@ -26,3 +26,8 @@ Note: In the below example, the group's front-end repository was named `bb-pract
 
 #### Running the front-end server in the browser
 ![browser running](images/front-end-running-browser.png)
+
+
+#### Formatting code
+ Prettier is installed in this repo and will auto-format files on git commits using the Husky tool.
+Prettier automatically analyzes your code and rewrites it according to a set of predefined rules. It takes care of indentation, line breaks, spacing, and other formatting aspects, ensuring that the resulting code is easy to read and understand.
```
akosasante: 👍 🤩

public/README.md
akosasante, line 1: This doc looks great! Currently it's in the `public` folder which is usually for assets (images, fonts, other files) that we want to have accessible from the website. In this case, I'm not sure that we want to do that for this README. I would either combine this info with the existing `README.md` in the root of the repository; or move this into a new folder called `docs` where all future documentation files for the repo could live.

.prettierrc.json
akosasante, line 1: For some reason this file is still showing up weird on Github (screenshot attached) but if it's showing correctly on your machine and running prettier then no worries. Just something to keep an eye on and make sure the other team members are able to get it working on their ends too.
Screenshot 2023-07-30 at 6 10 20 PM

akosasante (approved):
A comment about the placement of the README.md; otherwise looks good to me! Feel free to merge to main once you've taken a look at the comments and feel ready to merge.

public/README.md
JafariM: I will add a `docs` folder for future documentation files.

.prettierrc.json
JafariM: I don't know the reason, but it works good in my machine. I will ask others to check too

PR #2, opened 2023-07-31
update eslintConfig in package.json to solve the error

JafariM:


.eslintrc.json
akosasante, line 1: main thing here was just moving this to the root folder right?

akosasante (approved):
I have minimal knowledge of how eeslint operates in the react world; so I'm unsure if you need _both_ `react` and `react-app` in the config. But if it got things to work, then I'm happy to approve to get the fix into main :) :white\_check\_mark:

.eslintrc.json
JafariM: Yes, right

PR #3, opened 2023-07-31
Add tooling to document component props

JafariM:


nikkigraybeal (changes requested):
This looks good - it runs with no errors. I noticed there is an extra prettierrc.json file in the public folder. That needs to be removed. Also, you need to include node_modules in the prettierIgnore file. 
Nice work getting the error sorted and making a decision on using prop types, Masooma!!

JafariM:
Thank you for your review, I had a difficult time to set up prettier and enslint. I add some rules in  prettierrc.json file and as I read in an article there is no need to add node_modules to prettierIgnore file, because it is ignored by default.

nikkigraybeal:
That's great, Masooma! Thanks for letting me know about node_modules. I don't see any rules added to the prettierrc file - it's just an empty object. I think that's fine though if you and the team are ok with the default formatting it provides. Be sure to remove the extra prettierrc.json from the public folder though. I'm pretty sure there should only be one prettierrc.json file at the root of the project.

JafariM:
Sorry, I think the rules I add is in another pull request. There is only one  prettierrc.json file in my end, I couldn't see any extra in root folder

PR #5, opened 2023-08-03
MUI framework installed and skeleton form create

dayofthetech:
First PR after accept invitation.


package.json
```diff
@@ -3,6 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@emotion/react": "^11.11.1",
+    "@emotion/styled": "^11.11.0",
```
akosasante: Since this repo already has `styled-components` installed, and `emotion` is a simliar library, maybe we just stick with `styled-components`? If you have an inkling to try `emotion` instead, then that's totally fine, but we should not have both installed since it may eventually cause some styling conflicts.

**If we want to keep `styled-components`:**

[https://mui.com/material-ui/getting-started/installation/#with-styled-components](https://mui.com/material-ui/getting-started/installation/#with-styled-components)

```
# uninstall emotion libraries
npm rm @emotion/react @emotion/styled

# install material ui styled-components helper
npm i -S @mui/styled-engine-sc
```

**If we want to keep `emotion`:**

```
# uninstall styled-components
npm rm styled-components
```

src/components/Contact.jsx
akosasante, line 1: This looks great!

Will this component be used; or are you aiming to use it as a demo to show how to use the library?

I'm wondering if this PR can go ahead and wrap the `App.js` contents in the `ThemeProvider`, `Container`, and `CssBaseline` components so that subsequent PRs can start building with the outer layout already set?

akosasante (changes requested):
Looks great! I would encourage you to also include a short blurb in the `docs` folder so that the other frontend collaborators have a general idea of how to get started using Material UI. Can be an explanation of the library or links to any resources or docs that you think will be helpful.

package.json
dayofthetech: I think I'll keep styled components so I'll 

# uninstall emotion libraries
npm rm @emotion/react @emotion/styled

# install material ui styled-components helper
npm i -S @mui/styled-engine-sc

dayofthetech: I don't see the docs folder because when I cloned that was before it was created

src/components/Contact.jsx
dayofthetech: This is a demo sample

dayofthetech:
I believed I was able to solve for the merge conflicts

akosasante (commented):
Looking good! Your merge of main went well for the most part but I think it accidentally wiped away the changes in `package.json`. You can run the install steps again to install the MaterialUI and StyledComponents helper libraries, then commit and push that `package.json`. 

I think you should now also have the `docs` folder to write a blurb there. Or if you prefer you can do that as a follow up PR later 🙂

dayofthetech:
latest PR solves mui conflict

dayofthetech (commented):
mui

public/index.html
```diff
@@ -24,6 +24,10 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+    <!-- Material UI -->
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap"/>
+    <meta name="viewport" content="initial-scale=1, width=device-width" />
```
akosasante, line 29: I think this is already present on line 6

akosasante (approved):

PR #8, opened 2023-08-06
Create main.yml

dayofthetech:
Set up Github actions - Jira ticket

PR #10, opened 2023-08-08
Update main.yml

dayofthetech:
added three new github actions - 
prettier, lint, jest

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,15 @@ jobs:
       with:
         node-version: ${{ matrix.node-version}}
         cache: 'npm'
+
     - name: Installing dependencies
-      run: |
-        npm install
-        npm ci
+      run: npm install
+
+    - name: Running Prettier for formatting
+      run: npm run format:all
+
+    - name: Running ESLint for linting
+      run: npm run lint:check
+
+    - name: Running Jest for testing
+      run: npm test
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -16,7 +16,15 @@ jobs:
       with:
         node-version: ${{ matrix.node-version}}
         cache: 'npm'
+
     - name: Installing dependencies
-      run: |
-        npm install
-        npm ci
+      run: npm install
```
akosasante:
```suggestion
      run: npm ci
```

This should be `npm ci` : https://support.deploybot.com/article/131-why-developers-should-use-npm-ci-instead-of-npm-install-and-its-benefits

akosasante (commented):
Looks great!

Currently the action is failing because `npm test` fails: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/actions/runs/5791288142/job/15695864823?pr=10](https://github.com/Code-the-Dream-School/dd-prac-team4-front/actions/runs/5791288142/job/15695864823?pr=10)

You can take a look at the test in `src/App.test.js`. Currently the test is looking to find text on the main app page with the words "learn react".

But so far in this project we've changed what the text on that main page says. So we need to update the test to look for the right text. Once you do that, the Github Action should pass and then these PRs can be merged :slightly\_smiling\_face:

akosasante (commented):
Looks good! Now the only thing missing is fixing the failing test so that the Action will pass. Once that's done you can feel free to ping me and I'll approve this to merge. You probably don't need to ahve both this one and the other [PR open](https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/8/files) though since they're basically doing the same thing just in increments.

PR #12, opened 2023-08-12
Create a logout page

JafariM:


akosasante (commented):
@JafariM  Will you update this one to also use `react-auth-context` instead of `react-auth-kit` ?

src/App.js
```diff
@@ -1,19 +1,38 @@
...
+        (schema, request) => {
+          let attrs = JSON.parse(request.requestBody);
+          console.log(attrs);
+        }
+      );
+    //route to logout user
+    this.post('http://localhost:8000/api/v1/auth/logout', (schema, request) => {
+      let attrs = JSON.parse(request.requestBody);
+      console.log(attrs);
+    });
```
akosasante, line 26: After you merge your Register page PR, then this will have to move to the new `mirageServer` file

JafariM:
Yes, I will update it once the user register is merged

src/components/Logout.js
```diff
@@ -0,0 +1,41 @@
...
+        signOut(); // Log the user out from the client-side as well
+        //redirect user to home page
+        navigate('/home');
+      } else {
+        console.log('Logout failed');
+      }
+    } catch (error) {
+      console.error('Error during logout:', error);
+    }
+  };
```
akosasante, line 30: Will be able to use \`react-auth-context\` instead:

```suggestion
import { useLogout } from '@akosasante/react-auth-context';

const Logout = () => {  
 const navigate = useNavigate();
 const errorHandler = (error) => console.error("Error during logout: ", error);  
   
 const { submit: signOut, loading } = useLogout({ errorHandler, apiUrl: 'http://localhost:8000/api/v1/auth/logout' });  
   
 //handling user logout  
 const handleLogout = async () => {  
   const originalResponse = await signOut();  
   if (originalResponse.status === 201) {  
    navigate("/")  
   } else {  
      console.error("unexpected success response status when logging out")  
      // refresh page  
      navigate(0);  
   }  
 };  
}  
```

PR #14, opened 2023-08-15
github actions test

dayofthetech:
Rename component to SignIn 
Commented out the component because it was failing the test since SignIn component is not in `App.test.js` <- this can wait for now.

Added new line for main.yml to include 
`npm run test` atfer running this command it checks for one test and it passes, but I'm sure if command should check all three requierements, Prettier, ESLint and Jest (which I think is Craco)


akosasante (approved):
Great job Alberto! :+1: :rocket: 

> and Jest (which I think is Craco)

And yep just wanted to clarify, that `craco` [is](https://www.npmjs.com/package/@craco/craco) the replacement we're using instead of `react-scripts` [due](https://www.npmjs.com/package/react-scripts) to some issues we ran into earlier. They are both just script runners built on top of "Create React App" (CRA). CRA includes Jest built-in so when we run the test command it will be calling react-testing-library and Jest in the end.

akosasante:
Don't forget to move the ticket to "Done" column in Jira afterwards please.

PR #15, opened 2023-08-15
Create a register page

JafariM:


package.json
```diff
@@ -47,6 +50,7 @@
   },
   "devDependencies": {
     "@craco/craco": "^7.1.0",
+    "babel-preset-react-app": "@babel/plugin-proposal-private-property-in-object",
```
akosasante: Do you recall why this was needed?

public/index.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
+    <!-- <meta name="viewport" content="initial-scale=1, width=device-width" /> -->
```
akosasante: 👍 we can probably even just remove it completely since it's the same as line 7

src/App.js
```diff
@@ -29,7 +25,19 @@ const [message, setMessage] = useState('');
...
+        {/* only logged in user can visit home page */}
+        <Route
+          path="/home"
+          element={
+            <RequireAuth>
+              <Home />
+            </RequireAuth>
+          }
+        ></Route>
+      </Routes>
```
akosasante: Looks great!

Right now, if none of the routes are matched (so for example for localhost:8000, it won't match `/register`, `/signIn`, or `/home`) it will have a warning in the console:
```
No routes matched location "/" 
````

and it will just render whatever else this component has, which right now is just the message. We might want to default redirect users to the appropriate page as soon as the load the root url. Eg, we could base it on whether they are logged in or not:

```js
import { Routes, Route, Navigate } from 'react-router-dom';

function App() {
  const { status } = useAuth();

  return (
    <>
      <h1>{message}</h1>
      <Routes>
        <Route path="/" element={isLoggedIn ? <Home /> : <Navigate to="/signIn" /> }></Route>
        <Route path="/register" element={<UserRegistration />}></Route>
        <Route path="/signIn" element={<Contact />}></Route>
        <Route path="/home" element={<Home />}></Route>
      </Routes>
    </>
  );
}
```

akosasante (approved):
Looks great, feel free to merge; I've left a couple questions and suggestions but they can be resolved in a later ticket/PR if you prefer

package.json
JafariM: I don't remember why I need that here. I will remove it and test the app to see if any error happen or not.

PR #19, opened 2023-08-16
Create a navbar

JafariM:


src/App.js
```diff
@@ -25,14 +33,19 @@ function App() {
...
+        <Route
+          path="/"
+          element={isLoggedIn ? <Home /> : <Navigate to="/signIn" />}
+        ></Route>
         <Route path="/register" element={<UserRegistration />}></Route>
-        <Route path="/signIn" element={<Contact />}></Route>
+        <Route path="/signIn" element={<SignIn />}></Route>
         <Route path="/home" element={<Home />}></Route>
         <Route path="/logout" element={<Logout />}></Route>
+        <Route path="/logout" element={<Logout />}></Route>
```
nikkigraybeal, line 47: I see you're rendering the Logout component in the navbar as well. Is this just here for testing purposes? Either way, I think the Logout component can be removed from App.js.

src/components/Navbar.js
```diff
@@ -0,0 +1,32 @@
...
+const Navbar = () => {
+  const { status } = useAuth();
+  const isLoggedIn = status === AuthStatus.LoggedIn;
+  return (
+    <AppBar position="static">
+      <Toolbar>
+        <Typography variant="h6" component="div" sx={{ flexGrow: 1 }}>
+          Music
+        </Typography>
+        <Button color="inherit" component={Link} to="/signIn">
```
nikkigraybeal, line 16: I don't see a ticket for creating the navbar, so I'm not sure what needs to be implemented in this PR, but I would add conditional rendering to each link in this component based on whether the user is logged in or not. Logged in users should only see the logout and home links, while users who are not logged in should only see the signin and signup links. I see you already have the logout button being conditionally rendered! If there's a reason for not doing that for the rest of the links right now feel free to leave it as is.

src/util/mirageServer.js
nikkigraybeal, line 1: I've never used mirage.js before, so I'm not sure if your logout route is correct - sorry! It looks reasonable to me, but maybe @akosasante can double check for us to be sure?

nikkigraybeal (changes requested):
This PR looks good to me, Masooma! Great job! just a couple of very minor changes and I think it's good to go.

src/App.js
akosasante: I could be wrong, but I _think_ the way it's set up we still need it here (though right now it's duplicated so we can remove one of line 46 or 47) to define the route for `react-router-dom`. 

Maybe the routes should be declared elsewhere/earlier though to not make it confusing vs the actual content of the `App` component? I'm not 100% sure on best practices around that

src/util/mirageServer.js
akosasante: LGTM :+1:

Looks like for now, that mock route isn't returning anything just logging the request body (which I _think_ will be empty anyway). But that's fine as long as it allows testing the frontend without the backend (though in this case, I think this backend route is ready/merged to main so we could just pass that one through to localhost:8000).

src/components/Navbar.js
akosasante: +1

src/App.js
akosasante: ohh nvm I see; the component for `<Logout>` is just a button not a full page; so there's nothing really to route to. I agree then with @nikkigraybeal  we can remove line 46 and 47 here

akosasante:
Oh and after you merge in main; the navbar spacing issues _should_ be fixed automagically :slightly\_smiling\_face:

PR #20, opened 2023-08-16
Write documentation for testing

JafariM:


src/App.js
```diff
@@ -25,20 +33,18 @@ function App() {
...
-        />
-        <Route path="/logout" element={<Logout />} />
+          path="/"
+          element={isLoggedIn ? <Home /> : <Navigate to="/signIn" />}
+        ></Route>
+        <Route path="/register" element={<UserRegistration />}></Route>
+        <Route path="/signIn" element={<SignIn />}></Route>
+        <Route path="/home" element={<Home />}></Route>
+        <Route path="/logout" element={<Logout />}></Route>
+        <Route path="/logout" element={<Logout />}></Route>
```
akosasante, line 47: Per the navbar PR; I think we can remove these

src/App.js
```diff
@@ -1,16 +1,24 @@
...
 import Home from './components/Home';
+import { Routes, Route, Navigate } from 'react-router-dom';
 import Logout from './components/Logout';
-import SignIn from './components/SignIn';
-import { Routes, Route } from 'react-router-dom';
-import { RequireAuth } from '@akosasante/react-auth-context';
+import Navbar from './components/Navbar';
+import {
+  AuthStatus,
+  RequireAuth,
```
akosasante, line 11: I think we're no longer using RequireAuth?

src/App.js
```diff
@@ -1,16 +1,24 @@
...
+  useAuth,
+} from '@akosasante/react-auth-context';
 
 const URL = 'http://localhost:8000/api/v1/';
 
 function App() {
   const [message, setMessage] = useState('');
+  const { status } = useAuth();
+  const isLoggedIn = status === AuthStatus.LoggedIn;
+  // const navigate = useNavigate();
```
akosasante, line 21: We can remove this if it's not being used (and remove it from the import as well)

src/components/UserRegistration.js
```diff
@@ -60,14 +60,13 @@ const UserRegistration = () => {
   );
   // end of snackbar
 
-  const [signupError, setSignupError] = React.useState('false');
-  let errorMessage = '';
+  const [signupError, setSignupError] = React.useState(false);
+  const errorMessage = '';
```
akosasante, line 64: this should be `let` otherwise, you won't be able to reassign it on line 69

akosasante (approved):
docs look good; left comments about other stuff but those can probably be fixed in the appropriate PRs

PR #22, opened 2023-08-17
Updating app.js and user registration.js

JafariM:


src/App.js
```diff
@@ -4,21 +4,20 @@ import UserRegistration from './components/UserRegistration';
...
 
 function App() {
   const [message, setMessage] = useState('');
   const { status } = useAuth();
   const isLoggedIn = status === AuthStatus.LoggedIn;
-  // const navigate = useNavigate();
+  // create an axios instance
+  const instance = axios.create({
+    withCredentials: true,
+  });
```
akosasante, line 20: Instead of creating a new one, we should just import the one created in `src/apis/axiosClient.js`

src/apis/axiosClient.js
```diff
@@ -0,0 +1,18 @@
+import axios from 'axios';
+
+const axiosInstance = axios.create({
+  baseURL: 'http://localhost:8000/api/v1/auth',
```
akosasante, line 4:
```suggestion
  baseURL: 'http://localhost:8000/api/v1',
```

We will be making api calls to other endpoints as well like /api/v1/users, so we don't want all of them to start with `auth`

akosasante: Eg: take a look at the endpoints that have been documented so far: https://musicstore.onrender.com/api-docs/

src/util/mirageServer.js
```diff
@@ -35,6 +36,14 @@ export function makeServer() {
         }
       );
 
+      this.get(
+        'http://localhost:8000/api/v1/auth/showMe',
+        (schema, request) => {
+          let attrs = JSON.parse(request.requestBody);
+          console.log(attrs);
+        }
```
akosasante, line 44: This should return something so that when this mock is called it will return something as the `response.data`

src/components/Home.js
```diff
@@ -1,8 +1,27 @@
...
 
+  // testing axiosInstance request
+  const [data, setData] = useState([]);
+
+  useEffect(() => {
+    axiosInstance
+      .get('/showMe')
+      .then((response) => {
+        setData(response.data);
+        console.log('this is axios instance' + data);
```
akosasante, line 16: I wouldn't expect this to be the _axios instance_, I would expect this to be the HTTP response body returned by the axios instance.

akosasante (commented):
Just to make sure I'm on the same page; is this PR for a specific ticket, or just general cleanup?

JafariM:
> Just to make sure I'm on the same page; is this PR for a specific ticket, or just general cleanup?

Yes, it is only for cleanup

src/apis/axiosClient.js
JafariM: I removed `auth` from url

akosasante:
I'm not sure if this is still needed or can be closed?

JafariM:
> I'm not sure if this is still needed or can be closed?

I will close it

PR #24, opened 2023-08-23
Ps 54 create main storefront page

JafariM:


src/components/AlbulList.js
akosasante, line 1: There's a typo in the filename, should be `AlbumsList.js` . You may end up needing to change imports as well

src/components/AlbulList.js
```diff
@@ -0,0 +1,154 @@
...
+
+  return (
+    <Container>
+      {/* display snackbar if any error happened during API fetch */}
+      {hasError && (
+        <Snackbar
+          open={open}
+          autoHideDuration={6000}
+          onClose={handleClose}
+          message={errorMessage}
```
akosasante: I'm getting an `errorMessage is not defined` error because this variable isn't set up anywhere

src/App.js
akosasante, line 1: I don't think these changes are intentional? Or maybe you just need to pull in `main`?

src/components/AlbulList.js
akosasante: Maybe you'd want to have that in state, so something like:

```javascript
// at top of component

// then in the fetchAlbums function:
    } catch (error) {
      console.error('Error fetching albums:', error);
      setErrorMessage(error.message);
      setHasError(true);
    }
```

src/components/AlbulList.js
```diff
@@ -0,0 +1,154 @@
...
+      </IconButton>
+    </React.Fragment>
+  );
+
+  //end of snackbar
+
+  return (
+    <Container>
+      {/* display snackbar if any error happened during API fetch */}
+      {hasError && (
```
akosasante: We already control if the snackbar is open with the `open` variable. Now we also have to control `hasError` and make sure we set it back to false afterwards and all that. It's unecessary to have to juggle both of them, it should work with just the `open` variable:

```
  const [open, setOpen] = React.useState(false); // start snackbar as closed
  
  // inside fetchAlbums, when there's an error, then set the toast to opened. It will automatically close after the `autoHideDuration` so we don't have to handle setting open back to false afterwards ourselves
  
      } catch (error) {
      console.error('Error fetching albums:', error);
      setOpen(true);
      setErrorMessage(error.message);
    }
```

src/components/AlbulList.js
```diff
@@ -0,0 +1,154 @@
...
+        <Snackbar
+          open={open}
+          autoHideDuration={6000}
+          onClose={handleClose}
+          message={errorMessage}
+          action={action}
+        />
+      )}
```
akosasante:
```suggestion
        <Snackbar
          open={open}
          autoHideDuration={6000}
          onClose={handleClose}
          message={errorMessage}
          action={action}
        />
```

src/components/AlbulList.js
```diff
@@ -0,0 +1,154 @@
...
+    } catch (error) {
+      console.error('Error fetching albums:', error);
+      setHasError(true);
+    }
+  };
+
+  //call the function to make API request for search input value
+  const handleSearch = () => {
+    setLimit(1);
+    fetchAlbums(searchType, searchTerm, 1);
```
akosasante: Why are we doing `setLimit(1)` here? it hasn't changed right?

And we can pass in the limit in the `fetchAlbums` call instead of hardcoding it so that if we ever _do_ change the limit there's not too many places to edit:

```suggestion
    setLimit(1);
    fetchAlbums(searchType, searchTerm, limit);
```

src/components/Home.js
```diff
@@ -1,14 +1,35 @@
...
+  }, []);
+
+  // end of testing
+
   return (
-    <div>
+    <>
       <h1>THIS IS THE HOME PAGE</h1>
       <p>Current log-in status is: {status}</p>
       <p>Current logged-in user is: {JSON.stringify(user)}</p>
```
akosasante: We can probably remove this content at this point since we now have actually something to show here?

src/components/Home.js
```diff
@@ -1,14 +1,35 @@
...
+      .then((response) => {
+        setData(response.data);
+        console.log('this is axios instance' + data);
+      })
+      .catch((error) => {
+        console.error('Error fetching data:', error);
+      });
+  }, []);
+
+  // end of testing
```
akosasante: This testing stuff can be removed right? Might already be removed once you merge `main` into your feature branch

src/components/AlbulList.js
JafariM: Sorry, I updated it

src/components/Home.js
JafariM: Yes, I removed it

src/components/Home.js
JafariM: Removed

src/components/AlbulList.js
JafariM: Updated it

src/components/AlbulList.js
JafariM: Updated it

src/components/AlbulList.js
JafariM: You are right, that was extra work

akosasante (commented):
Alberto's ticket has been merged to main now; so when you do a `git pull origin main` you will probably have some conflicts to resolve. And then you can replace your:

```
{albums.length > 0 ? (
        <>{/* <SearchResult searchResult={albums} /> */}</>
      )
```

with

```
{albums.length > 0 ? (
        <AlbumGrid albums={albums} />
      )
```

and then you can see the results show up :yay:!

src/components/AlbulList.js
akosasante: I meant to also share this link to explain a bit about why to use state instead of a plain variable in this case. Basically anytime you're wanting something in the UI to change as a result of changing a variable then it should probably be state or at least derived from state. Otherwise, plain variables are fine. https://stackoverflow.com/questions/58252454/react-hooks-using-usestate-vs-just-variables#:~:text=That%20means%2C%20when%20the%20value,useState%20than%20normal%20local%20value.&text=Local%20vaiables%20value%20doesn't,the%20variable%20value%20between%20renders.

src/components/AlbumsList.js
```diff
@@ -0,0 +1,152 @@
...
+        </Grid>
+        <Grid item xs={12} sm={2}>
+          <Button variant="contained" onClick={handleClear} fullWidth>
+            Clear
+          </Button>
+        </Grid>
+      </Grid>
+
+      {albums.length > 0 ? (
+        <>{/* <SearchResult searchResult={albums} /> */}</>
```
akosasante: Here is where we can add `<AlbumGrid>`

src/App.js
akosasante, line 1: Some changes here that I don't expect; we should pull in main; and hopefully these changes will go away

akosasante (approved):
A couple small comments but overall looking good to go for me. You may just want to do `git pull origin main` and `npm install` to try and get the Gihub Actions CI passing

src/components/AlbumsList.js
```diff
@@ -0,0 +1,152 @@
...
+      if (searchResult.length === 0) {
+        setMessage('No result found');
+      }
+    } catch (error) {
+      console.error('Error fetching albums:', error);
+      setErrorMessage(error.message);
+      setOpen(true);
+    }
+  };
+
```
akosasante, line 51: Oh I also meant to add earlier; it would be nice to just have some initial values shown, when the user comes to this screen without having them have to press search first. We could do that by calling `fetchAlbums` as soon as this component is loaded, with no parameters. We can use `useEffect` to do something when the component mounts, by passing in an empty array as the second argument: https://medium.com/devil-is-in-the-details/dependency-array-in-useeffect-hook-d73e0ef2ab33

```js
useEffect(() => {
  fetchAlbums('albumName', '', 10)
}, []);

JafariM: Yes, that is something I wanted to add too, but I was waiting for AlbumGrid to merge first. I will add that

PR #25, opened 2023-08-23
Loader implenented

AlonaVladymyrovaTrinity:
Loader component successfully created.
This is functional styled 'Loader' component that displays a spinner with optional size variations based on the given class name.

Testing Instructions:

Install the necessary dependencies by running npm install in the project directory.
Signup to the application 
Navigate to Home page http://localhost:3000/ 
Verify that the Home page displays a loader


akosasante (approved):
Looks great!; I'm guessing we'll just comment out/remove it from Home.js before merging?

AlonaVladymyrovaTrinity:
Yes, of course, I added this component to the Home page only for review.
I will remove it from homepage before merge.

PR #26, opened 2023-08-23
add authentication to login page using react-auth-context

JafariM:


src/components/SignIn.js
```diff
@@ -10,6 +10,12 @@ import Box from '@mui/material/Box';
 import LockOutlinedIcon from '@mui/icons-material/LockOutlined';
 import Typography from '@mui/material/Typography';
 import Container from '@mui/material/Container';
+import { useRegister } from '@akosasante/react-auth-context'; // authentication library
```
akosasante: This is for logging in right? so we just need `useLogin` not both `useLogin` and `useRegister`

src/components/SignIn.js
```diff
@@ -30,17 +36,85 @@ function Copyright(props) {
...
+  const [formData, setFormData] = useState({
+    email: '',
+    password: '',
+  });
+
+  //update the input states when their value is changed
+  const handleChange = (e) => {
+    const { name, value } = e.target;
+    setFormData((prevData) => ({ ...prevData, [name]: value }));
+  };
```
akosasante: 🤩  nice job!

src/components/SignIn.js
```diff
@@ -30,17 +36,85 @@ function Copyright(props) {
...
-      email: data.get('email'),
-      password: data.get('password'),
-    });
+    setLoginError(false); // Reset login error state
+    const originalResponse = await signIn();
+    console.dir(originalResponse);
+    //navigate to home page only if credential is correct
+    if (originalResponse.status !== 401) {
+      navigate('/home');
+    }
```
akosasante: If the request is unsuccessful, the error will be handled by `handleLoginError`. We're not currently returning anything from there, so when there's an error `originalResponse` will be undefined. Which means this will throw an error because you can't do `undefined.status`.

Easiest thing here would probably be to check for the _success_ case and use optional chaining (`?.` to handle the null/undefined case):

```suggestion
    if (originalResponse?.status === 201) {
      navigate('/home');
    }
```

src/components/SignIn.js
```diff
@@ -55,6 +129,16 @@ export default function SignIn() {
...
+        {/* display snackbar if any error happened during user login */}
+        {loginError && (
+          <Snackbar
+            open={open}
+            autoHideDuration={6000}
+            onClose={handleClose}
+            message={errorMessage}
+            action={action}
+          />
+        )}
```
akosasante: The Snackbar component handles by itself internally whether to be displayed or not (based on things like the `autoHideDuration` that you give it. So we should not need to do further conditional rendering logic, and in fact it may lead to sneaky bugs:

```suggestion
        {/* display snackbar if any error happened during user login */}
          <Snackbar
            open={open}
            autoHideDuration={6000}
            onClose={handleClose}
            message={errorMessage}
            action={action}
          />
```

We just need to make sure we only set `open` to `true` when we want to display the Snackbar (eg: in the `handleLoginError` function)

src/components/SignIn.js
```diff
@@ -95,14 +182,23 @@ export default function SignIn() {
...
+      {/* useLogin hook errors */}
+      {errors && (
+        <Snackbar
+          open={open}
+          autoHideDuration={6000}
+          onClose={handleClose}
+          message={errors.msg} // Use the error message from the errors object
+          action={action}
+        />
+      )}
```
akosasante: Is there a reason we have two Snackbar components in this container?

src/components/SignIn.js
JafariM: I add that condition because even with wrong credential user was navigated to home page, I will update it with your code

src/components/SignIn.js
JafariM: Sorry, I copied the code to use the path for auth library and forgot to delete it

src/components/SignIn.js
JafariM: You are right, I missed that point

src/components/SignIn.js
JafariM: Maybe I got it wrong at time, I thought because we have 2 errors one from useLogin hook another from handleLoginError function so the message will be different. I created 2 snackbar for that

src/components/SignIn.js
```diff
@@ -30,17 +35,85 @@ function Copyright(props) {
...
+  //update the input states when their value is changed
+  const handleChange = (e) => {
+    const { name, value } = e.target;
+    setFormData((prevData) => ({ ...prevData, [name]: value }));
+  };
+
+  const navigate = useNavigate();
+
+  // snackbar
+  const [open, setOpen] = React.useState(true);
```
akosasante:
```suggestion
  const [open, setOpen] = React.useState(false);
```

Otherwise, we're starting with the snackbar open but empty:

src/components/SignIn.js
```diff
@@ -10,6 +10,11 @@ import Box from '@mui/material/Box';
 import LockOutlinedIcon from '@mui/icons-material/LockOutlined';
 import Typography from '@mui/material/Typography';
 import Container from '@mui/material/Container';
+import { useNavigate } from 'react-router-dom';
+import Snackbar from '@mui/material/Snackbar';
+import IconButton from '@mui/material/IconButton';
+import CloseIcon from '@mui/icons-material/Close';
+import { useLogin } from '@akosasante/react-auth-context';
 
 function Copyright(props) {
```
akosasante: We can delete this function now that we're not using it

src/components/SignIn.js
akosasante: The one we process in the `handleLoginError` function is the same as what's returned from the `useLogin` hook. We just use that `handleLoginError` function so we can do more steps like, opening the snackbar and logging the error.  You can try it by doing a `console.log("Error from handleLoginError: ", error)` inside `handleLoginError` and then a `console.log("Error from hook: ", error)` under the `useLogin` Hook. 

In fact, in this case we can probably just leave line 96 as: `const { submit: signIn, loading, } = useLogin(formData, loginHookOptions);` since we don't need to use the raw errors directly in our code here. 

Also material-ui doesn't really allow two snackbars to be shown at the same time, one will overtake the other (though in this case it wouldn't have been noticeable since both snackbars would've had the same content anyway) : https://mui.com/material-ui/react-snackbar/#consecutive-snackbars

akosasante: An alternative option; would be to keep just the second snackbar, remove the `errors &&`, and make it only open when the errors is not null:

```
<Snackbar
  open={!!errors}
  autoHideDuration={6000}
  onClose={handleClose}
  message={errors?.msg}
  action={action}
/>
```

In this option, we would need to keep what you have currently where we get the errors from the hook's internal state: `const { submit: signIn, loading, errors } = useLogin(formData, loginHookOptions);`

And we would only use `handleLoginError` for logging but _not_ for setting any kind of UI state / showing/hiding the snackbar etc.

src/components/SignIn.js
```diff
@@ -30,17 +35,85 @@ function Copyright(props) {
...
+        onClick={handleClose}
+      >
+        <CloseIcon fontSize="small" />
+      </IconButton>
+    </React.Fragment>
+  );
+  // end of snackbar
+
+  // user authentication
+  const [loginError, setLoginError] = React.useState(false);
```
akosasante: `loginError` is no longer used so we can remove this and all the places we're calling `setLoginError`

src/components/SignIn.js
```diff
@@ -30,17 +35,85 @@ function Copyright(props) {
...
+  let errorMessage = '';
```
akosasante: Per the other PR, if we will be using this to render things in the UI, then we should be setting it with `useState`

src/components/SignIn.js
JafariM: Thank you for clarification, I update it

src/components/SignIn.js
JafariM: I deleted it

src/App.js
```diff
@@ -1,29 +1,15 @@
...
 import SignIn from './components/SignIn';
 import Home from './components/Home';
 import { Routes, Route, Navigate } from 'react-router-dom';
 import Navbar from './components/Navbar';
 import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
 
-const URL = 'http://localhost:8000/api/v1/';
-
 function App() {
   const [message, setMessage] = useState('');
```
akosasante: Sorry, one more thing we can probably remove; Along wiht line 16 where `message` is used.

akosasante (commented):
Since the other branch got merged in; you'll probably want to do a git pull from your local branch of this branch and fix merge conflicts so that you don't lose your previous changes here.

src/App.js
JafariM: Removed.

src/components/SignIn.js
akosasante: @JafariM I think this still applies; maybe you fixed it before but it looks like it's still there, maybe it got undone with the last merge or something?

src/components/SignIn.js
akosasante: @JafariM I think this still applies; maybe you fixed it before but it looks like it's still there, maybe it got undone with the last merge or something?

src/components/SignIn.js
akosasante: @JafariM I think this still applies; maybe you fixed it before but it looks like it's still there, maybe it got undone with the last merge or something?

src/components/SignIn.js
akosasante: [@JafariM](https://github.com/JafariM) I think this still applies; maybe you fixed it before but it looks like it's still there, maybe it got undone with the last merge or something?

akosasante:
Once you've removed the code that used to show "This is a music app" on the front page, the test in `App.test.js` will probably start to fail. You can change it to instead look for `/Music/` since show that text in the navbar

src/components/SignIn.js
JafariM: > 

Yes it was undone, I just delete them agian

src/components/SignIn.js
```diff
@@ -72,13 +57,9 @@ export default function SignIn() {
   // end of snackbar
 
   // user authentication
-  const [loginError, setLoginError] = React.useState(false);
-  let errorMessage = '';
 
   // error handler function
   const handleLoginError = (error) => {
```
akosasante, line 62: we still need something to open the snackbar when there's an erorr. there's a few ways we can do this. Option 1 is to open the snackbar manually by changing the `open` state here in the error handler where we know an error has occured:

```
const handleLoginError = (error) => {
  setOpen(true);
}
```

src/components/SignIn.js
```diff
@@ -189,15 +160,13 @@ export default function SignIn() {
...
-        <Snackbar
-          open={open}
-          autoHideDuration={6000}
-          onClose={handleClose}
-          message={errors.msg} // Use the error message from the errors object
-          action={action}
-        />
-      )}
+      <Snackbar
+        open={open}
```
akosasante: Option 2 for opening this on an error is to just make the Snackbar.open prop be looking at whether the `errors` variable from the hook is populated or not:

```
<Snackbar
  open={!!errors}
  autoHideDuration={6000}
  onClose={handleClose}
  message={errors?.msg} // Use the error message from the errors object
  action={action}
 />
```

src/components/SignIn.js
```diff
@@ -189,15 +160,13 @@ export default function SignIn() {
...
+        autoHideDuration={6000}
+        onClose={handleClose}
+        message={errors.msg} // Use the error message from the errors object
```
akosasante: Since `errors` can be null (ie: when you first load the page, there's no errors yet) this is currently causing an error when you try to load the page



This error is telling us that Javascript is trying to access `errors.msg` but it can't because `errors` is null.

We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here to tell Javascript that in some cases this value will be null, in which case just return null instead of trying to access `.msg`: 

```
message={errors?.msg}
```

akosasante (approved):
A couple small fixes and then make sure all the formatting/linting is working and you can merge

src/components/SignIn.js
JafariM: > Option 2 for opening this on an error is to just make the Snackbar.open prop be looking at whether the `errors` variable from the hook is populated or not:
> 
> ```
> <Snackbar
>   open={!!errors}
>   autoHideDuration={6000}
>   onClose={handleClose}
>   message={errors?.msg} // Use the error message from the errors object
>   action={action}
>  />
> ```

Sorry, I did this before, but it did undone and I forgot to update this part too.

PR #27, opened 2023-08-24
PS-85 Create a checkout page

AlonaVladymyrovaTrinity:
**Checkout page created**

**Pull Request Description:**

This pull request contains next new following files: 
src/components/Purchase/CheckoutForm.js
src/components/Purchase/CheckoutForm.module.css
src/components/Purchase/CheckoutPage.js
src/components/Purchase/CheckoutPage.module.css

src/components/Purchase/CheckoutPage.js: This file contains a React component that represents the checkout process page. It retrieves the client secret for the payment intent from the server based on the payment amount, and then renders the CheckoutForm component with the appropriate options.

src/components/Purchase/CheckoutForm.js: This file contains a React component that represents the checkout form for making a payment. It handles the Stripe payment integration and displays the payment element and relevant messages based on the payment status.

**Please ensure that the necessary environment variables(like Stripe secret key) and dependencies are properly set up before testing the pull request.**

**Testing Steps:**

- Install the required dependencies using npm install.
Install:

```
npm install 
npm install @stripe/react-stripe-js @stripe/stripe-js
```

- **Make sure the Stripe API keys are properly configured in the environment variables.**
- Run the application locally using npm start.
- Open your browser and access the checkout page.

[http://localhost:3000/checkout](http://localhost:3000/checkout)

- Submit the payment form and verify that the validation messages are displayed correctly for invalid inputs.
- If the input is valid, the checkout process page should be loaded.
- Verify that the checkout form is displayed and functional.
- Test the Stripe integration by entering valid **testing payment details** and completing the payment.

> **Testing payment details:**
> Email: any
> Card number: 
> 4242 4242 4242 4242
> Expiration: any(but date should be in future)
> CVC: any
> Country: United States
> Zip: any

- Test different scenarios, such as successful payment, processing payment, and failed payment, to ensure proper handling of different payment statuses.
- Test the overall flow of the checkout process from start to finish, ensuring that the data is accurately processed and the user is redirected appropriately.

Screenshot 2023-08-23 at 9 17 42 PM


src/components/Purchase/CheckoutPage.js
```diff
@@ -0,0 +1,51 @@
...
+  'pk_test_51Nfj2pKnb0YvaiB3W0z31bbB7OPEnqem6wnpSbohudeiDkj1NUde9M5kgUwozzrfBqGqRpWU5ivDIInzWdt4q5zw00vMz7fc1c'
+);
+const paymentAmount = 1099; // This is $10.99 payment amount(added for testing) This amount variable should came from cart
+
+const CheckoutPage = () => {
+  const [clientSecret, setClientSecret] = useState('');
+
+  useEffect(() => {
+    // Create PaymentIntent as soon as the page loads
+    fetch('http://localhost:8000/api/v1/stripe', {
```
akosasante: Where is this URL defined? 

I think we actually want to call the order create endpoint in the backend: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/orderController.js#L6](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/controllers/orderController.js#L6)

It's at `http://localhost:8000/api/v1/orders`.

I don't think it's documented yet so please check with the team what the expected input/output is. But I know it will take some order items, and return the created order + clientSecret.

src/components/Purchase/CheckoutForm.js
```diff
@@ -0,0 +1,116 @@
...
+          break;
+        case 'requires_payment_method':
+          setMessage('Your payment was not successful, please try again.');
+          break;
+        default:
+          setMessage('Something went wrong.');
+          break;
+      }
+    });
+  }, [stripe]);
```
akosasante: I don't believe we actually get to see these messages (at least form my testing), cause when the stripe `confirmPayment` function completes, it will immediately redirect to the checkout completed URL. That's likely where we'll want to be showing this kind of messaging about whether the payment succeeded or not.

The frontend team met recently to discuss what this may look like: [https://www.figma.com/file/ewQrLWdJH9xkqBuHnqSjER/Wireframing-(Copy)?type=design&node-id=766671-157&mode=design&t=2xrO7nisHMdEZWPQ-4](https://www.figma.com/file/ewQrLWdJH9xkqBuHnqSjER/Wireframing-(Copy)?type=design&node-id=766671-157&mode=design&t=2xrO7nisHMdEZWPQ-4)

It will be part of the code to be written for the next ticket: https://codethedreamteam4-2023.atlassian.net/browse/PS-95

src/components/Purchase/CheckoutForm.js
```diff
@@ -0,0 +1,116 @@
...
+      onSubmit={handleSubmit}
+      className={style.checkoutFormContainer}
+    >
+      <p>
+        Your payment amount: <b>${paymentAmount / 100}</b>
+      </p>
+      <LinkAuthenticationElement
+        id="link-authentication-element"
+        onChange={(e) => setEmail(e.target.value)}
+      />
```
akosasante: This component's `onChange` doesn't return a normal event object with a target field. So this is currently causing some errors that you can see in the console when you type into the input box generated by this element:



If you set a breakpoint on this line, you'll see that the shape of the argument `e` is actually a bit different than we expect, and the email is actually under `e.value` not `e.target.value`:



---

for bonus insight, if we dig into this component's props in the react-stripe-js library, we can see they're defined here in a typescript file:

[https://github.com/stripe/react-stripe-js/blob/b516a5e44c03de91093162acf1b33d89a7fbc446/src/types/index.ts#L334-L340](https://github.com/stripe/react-stripe-js/blob/b516a5e44c03de91093162acf1b33d89a7fbc446/src/types/index.ts#L334-L340)

which is then further defiend here in the stripe-js library:

[https://github.com/stripe/stripe-js/blob/4ba4eadc8f14040f9dc0bf916b9c09c383f6d603/types/stripe-js/elements/link-authentication.d.ts#L136-L158](https://github.com/stripe/stripe-js/blob/4ba4eadc8f14040f9dc0bf916b9c09c383f6d603/types/stripe-js/elements/link-authentication.d.ts#L136-L158)

Which matches the fields we saw when we did the breakpoint

akosasante (changes requested):

src/components/Purchase/CheckoutPage.js
akosasante: As well, that endpoint requires a cookie because the backend needs to authenticate who the user is who's making the order and that they've logged in. 

With `fetch` you can send cookies by adding `credentials: 'include'` to your options: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#sending_a_request_with_credentials_included

If you don't mind though I'd ask if you can convert this from using `fetch` to using the axios client defined in the `src/apis` folder to follow the pattern/standard that we'll be trying to use in this repo. More info here: https://codethedreamteam4-2023.atlassian.net/wiki/spaces/PT4/pages/5734401/Making+API+calls+from+the+frontend

src/components/Purchase/CheckoutForm.js
```diff
@@ -0,0 +1,99 @@
...
+    // Stripe.js has loaded
+    setIsStripeLoaded(true);
+    const clientSecret = new URLSearchParams(window.location.search).get(
+      'payment_intent_client_secret'
+    );
+
+    if (!clientSecret) {
+      return;
+    }
+  }, [stripe]);
```
akosasante, line 31: It's not clear to me what this `useEffect` is used for here

src/components/Purchase/CheckoutPage.js
AlonaVladymyrovaTrinity: I did changed the endpoint to http://localhost:8000/api/v1/orders, thanks.
And thank you for pointing me to Cookies issue. Which exact Cookies I should include in my callout?
Sure, I will try to convert fetch to axios after I will resolve authorizations issues.

src/components/Purchase/CheckoutForm.js
AlonaVladymyrovaTrinity: I removed this part with error message handling here.

src/components/Purchase/CheckoutForm.js
AlonaVladymyrovaTrinity: Thank you! Good point! I fixed this!

src/components/Purchase/CheckoutPage.js
AlonaVladymyrovaTrinity: I have included credentials: 'include' to my API call.

AlonaVladymyrovaTrinity: I added order info to my API request, but the code still doesn't work. What other changes do I need to make? Can you please advise something?

src/components/Purchase/CheckoutPage.js
```diff
@@ -0,0 +1,63 @@
...
+
+// Make sure to call loadStripe outside of a component’s render to avoid
+// recreating the Stripe object on every render.
+// This is your test publishable API key.
+const stripePromise = loadStripe(
+  'pk_test_51Nfj2pKnb0YvaiB3W0z31bbB7OPEnqem6wnpSbohudeiDkj1NUde9M5kgUwozzrfBqGqRpWU5ivDIInzWdt4q5zw00vMz7fc1c'
+);
+//Temporary variables for testing
+//const paymentAmount = 1099; // This is $10.99 payment amount(added for testing) This amount variable should came from cart
+const album = '876r75675';
```
akosasante: This is currently causing an error because its an invalid id for mongo (too short to be a valid Mongo ObjectId).

We can grab one of the existing album ids from `localhost:8000/api/v1/albums`  and use that as a placeholder here:

```suggestion
const album = '64d2a94c793389a43fc5a8d6';
```

akosasante (approved):
One note about the id used but after that feel free to merge, great job!

src/components/Purchase/CheckoutForm.js
AlonaVladymyrovaTrinity: That's a great question, thanks for asking.
This is important because a PaymentIntent needs to be created on the server before initiating the payment process with Stripe.
By using useEffect with [subtotal] as its dependency array, you ensure that the PaymentIntent is fetched and the clientSecret is updated whenever the subtotal value changes. This is useful in scenarios where the subtotal is dynamic and can change during the user's interaction with the page, and you want to ensure that the latest clientSecret is used when initiating the payment.

src/components/Purchase/CheckoutPage.js
AlonaVladymyrovaTrinity: I successfully updated the album ID.

src/components/Purchase/CheckoutForm.js
akosasante: Ah I see yeah that's a good idea :+1: 

Here it's using `stripe` as the dependency array though? or are you saying that as we build this out with the real data, the idea would be that we'd also have subtotal in there?

src/components/Purchase/CheckoutPage.js
AlonaVladymyrovaTrinity: The problem resolved. The issue was related to variable name in .env

PR #28, opened 2023-08-24
Ps 114 hide login and register in navbar

JafariM:


src/components/SignIn.js
akosasante, line 1: I'm not sure if these changes are meant to be in this PR.

src/util/mirageServer.js
```diff
@@ -15,6 +15,38 @@ export function makeServer() {
       // <-- FIX
 
       this.get('http://localhost:8000/api/v1', { data: 'This is a music app' }),
+        // route to submit user registration form
```
akosasante, line 18: Maybe add a little comment to the top of this file directing developers to uncomment any routes that they need/want to mock

akosasante (approved):
Navbar and mirageServer changes look good. Probably don't want the SignIn changes here since they're being reviewed in a separate PR.

src/util/mirageServer.js
JafariM: I will add the comment

PR #29, opened 2023-08-25
Ps 116 adding .env file

JafariM:


src/App.js
```diff
@@ -7,23 +7,10 @@ import { Navigate, Route, Routes } from 'react-router-dom';
 import Navbar from './components/Navbar';
 import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
 
-const URL = 'http://localhost:8000/api/v1/';
-
 function App() {
   const [message, setMessage] = useState('');
```
akosasante, line 10: We can remove this as well, it's no longer used here

src/App.js
```diff
@@ -7,23 +7,10 @@ import { Navigate, Route, Routes } from 'react-router-dom';
 import Navbar from './components/Navbar';
 import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
```
akosasante, line 8: We can also remove `useEffect` and `getAllData` from the imports as they're not used anymore

akosasante (approved):
LGTM! :+1: 

I changed the base branch to the `connect login` one because this PR seemed to include commits from there. Since we don't want to accidentally merge those until that branch is fixed, I changed the base branch to reflect that for now

src/App.js
JafariM: I removed them

JafariM:
> LGTM! 👍
> 
> I changed the base branch to the `connect login` one because this PR seemed to include commits from there. Since we don't want to accidentally merge those until that branch is fixed, I changed the base branch to reflect that for now

Good, thank you

PR #31, opened 2023-08-28
Ps 95 show the results on a checkout complete page

AlonaVladymyrovaTrinity:
**Pull Request Description**

**Changes Made:**

I have implemented and made changes to several React components to enhance the checkout process. These changes include handling Stripe payment status, loading order details, and displaying relevant information to the user.

**Testing Steps:**

1. Install dependencies:

   
`   npm install`
  

2. Run application:

 Front-end: `  npm start`
 Back-end: `npm run dev`

3. Open the application in your web browser.

4. Navigate to the checkout page and complete a payment transaction.

[http://localhost:3000/checkout](http://localhost:3000/checkout)

5. Observe the following:

   - Ensure that the payment status is displayed correctly.
   - Confirm that the order details are loaded and displayed when payment is successful.
   - Check for error messages in case of payment failure.

7. Verify that the "Return to Marketplace" button works correctly and redirects to the desired page.

Screenshot 2023-09-05 at 12 31 16 AM


src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+
+const stripePromise = loadStripe(
+  'pk_test_51Nfj2pKnb0YvaiB3W0z31bbB7OPEnqem6wnpSbohudeiDkj1NUde9M5kgUwozzrfBqGqRpWU5ivDIInzWdt4q5zw00vMz7fc1c'
+);
+
+const CheckoutComplete = () => {
+  //---This is a temporarily data for testing purpose---
+  let n = 1;
+  //----------------------------------------------------
+  const location = useLocation();
```
akosasante: This is fine as-is but wanted to share that react-router has a dedicated hook for this (since v6 I believe). So you could do:

```javascript
const [searchParams, setSearchParams] = useSearchParams();
const clientSecret = searchParams.get("payment_intent_client_secret")
```

And then have your `useEffect` be dependent on `[clientSecret]` instead of `[location]`

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+          `http://localhost:8000/api/v1/orders/${id}`,
+          {
+            withCredentials: true,
+            headers: {
+              'Content-Type': 'application/json',
+            },
+          }
+        );
+        setIsLoading(false);
+        setOrderData(response.data);
```
akosasante: This endpoint returns an object like:

```json
{
  "order": {
    "_id": "64f0ea30375f230a5e1832bb",
    "user": {
      "_id": "64ef5e90086e1c19b888c72c",
      "name": "Akos",
      "username": "akos",
      "email": "akos123456@example.com",
      "role": "admin",
      "__v": 0,
      "id": "64ef5e90086e1c19b888c72c"
    },
    "orderStatus": "complete",
    "tax": 0.15,
    "subtotal": 500,
    "total": 550,
    "orderItems": [
      {
        "album": "64d36b781a3159bca4e17193",
        "quantity": 5,
        "_id": "64f0ea30375f230a5e1832bc"
      }
    ],
    "createdAt": "2023-08-31T19:29:52.413Z",
    "updatedAt": "2023-08-31T19:29:52.891Z",
    "__v": 0,
    "paymentIntentId": "pi_3NlGYuKnb0YvaiB30vxdML9B"
  }
}
```

so we need to either make sure that all the places in this code that are accessing `orderData` object, do so by doing `orderData.order.<whatever>` or we can here do `setOrderData(response.data?.order)`

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+  useEffect(() => {
+    const queryParams = new URLSearchParams(location.search);
+    const clientSecret = queryParams.get('payment_intent_client_secret');
+    setClientSecret(clientSecret);
+
+    const orderDetails = async (id) => {
+      setIsLoading(true);
+      setErrorMessage('');
+      try {
+        const response = await axios.get(
```
akosasante: Recommend using the base axios client from the `apis` folder; as that will automatically set some common logic for you

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+
+  // Callback function to update isPaymentSuccessful
+  const updatePaymentStatus = (isSuccess) => {
+    setIsPaymentSuccessful(isSuccess);
+  };
+
+  useEffect(() => {
+    const queryParams = new URLSearchParams(location.search);
+    const clientSecret = queryParams.get('payment_intent_client_secret');
+    setClientSecret(clientSecret);
```
akosasante: If we use `useSearchParams` then these three lines won;t be needed

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+        setIsLoading(false);
+        setErrorMessage(
+          `An error occurred during loading order details with id ${id}`
+        );
+      }
+    };
+    orderDetails('64e950728af0bef9f0ced812');
+  }, [location]);
+
+  return (
```
akosasante, line 58: The return value here is really nested with a lot of conditional rendering logic which can make it difficult to debug. 

It might help to extract come code into subcomponents. These could live in the same file or perhaps create a `CheckoutComplete` subdirectory, since they're unlikely to be re-used outside of this component.

Eg:  the header content could be extracted so that this return starts with:

```js
    <>
      <Container className={style.mainContainer}>
        <Grid container spacing={3}>
          <Grid item xs={12} sm={6}>
            <CheckoutMessageCard orderData={orderData} clientSecret={clientSecret} updatePaymentStatus={updatePaymentStatus} isPaymentSuccessful={isPaymentSuccessful} isLoading={isLoading} errorMessage={errorMessage} userEmail={userEmail} />
          </Grid>
```

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+                      updateStatus={updatePaymentStatus}
+                    />
+                  </Elements>
+                </Typography>
+              </Box>
+              {isPaymentSuccessful ? (
+                <>
+                  {isLoading ? (
+                    <Loader className="small-spinner" />
+                  ) : errorMessage !== '' ? (
```
akosasante: The nested ternary can be unnested somewhat with `&&` boolean short circuiting: https://blog.logrocket.com/react-conditional-rendering-9-methods#short-circuit-and-operator

```js
{!isPaymentSuccessful && (
                <Typography variant="h6">
                  Payment was unsuccessful! Sorry for the inconvenience.
                </Typography>
              )}

{isPaymentSuccessful && (
                <>
                  {isLoading && (
                    <Loader className="small-spinner" />
                  )}
                  {!isLoading && !!errorMessage !== (
                    <Typography variant="h6">{errorMessage}</Typography> // Other order data fields can be added here
                  )}
                  {!isLoading && !errorMessage && (
                    <>
                      <Typography variant="h6">Order Details:</Typography>
                      <Typography variant="h6">
                        Order ID: {orderData.id}
                      </Typography>
                      <Typography>Amount: {orderData.total}</Typography>
                    </>
                  )}
                </>
              )}
```

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+                    <>
+                      <Typography variant="h6">Order Details:</Typography>
+                      {/* <Typography variant="h6">
+                        Order ID: {orderData.id}
+                      </Typography>
+                      <Typography>Amount: ${orderData.amount}</Typography> */}
```
akosasante: the order total is in the `total` property, rather than `.amount`

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+            <Paper className={`${style.paper} ${style.rightPaper}`}>
+              {isPaymentSuccessful ? (
+                <>
+                  {isLoading ? (
+                    <Loader className="small-spinner" />
+                  ) : errorMessage !== '' ? (
+                    <>
+                      <Typography variant="h6">Order Details:</Typography>
+                      {/* <Typography variant="h6">
+                        Order ID: {orderData.id}
```
akosasante: The order id will be under the `_id` property

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,146 @@
...
+              </Box>
+              {isPaymentSuccessful ? (
+                <>
+                  {isLoading ? (
+                    <Loader className="small-spinner" />
+                  ) : errorMessage !== '' ? (
+                    <>
+                      <Box mt="2rem">
+                        <Typography variant="h8">
+                          {/* {`Your order of ${orderData.order.orderItems.quantity} albums has been completed.`} */}
```
akosasante: The `orderItems` property will be a list of items: `[ {album: <album id>, quantity: <integer>, user: <user id> } ]`. The number of distinct albums ordered will be length of that list. Or if you want to indicate how many total items were purchased you may need to iterate through the list and add up all the quantities

akosasante:
@AlonaVladymyrovaTrinity  The tickets to add additional metadata to the paymentIntent and to the order object have been merged to main. So there's some things we can do now.

- In `CheckoutPage` you create the order, at that point you have access to the created order: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/src/components/Purchase/CheckoutPage.js#L27-L51 
	- You could pass that down via props to `CheckoutForm` and then include the order id in the `return_url` that Stripe uses to redirect to the `CheckoutComplete` page. (eg: `return_url: ${window.location.origin}/checkout/completed?orderId=${order._id}`)
	- When `CheckoutComplete` loads it would have a URL like `completed?orderId=64f1ed3274d34819a7c44de1&payment_intent=pi_3NlXoNKnb0YvaiB31DdrCiiy&payment_intent_client_secret=pi_3NlXoNKnb0YvaiB31DdrCiiy_secret_Tcv83sm5hfUfCO3DLQVOVXr63&redirect_status=succeeded` so you could use `useSearchParams` hook (or native `URLSearchParams`) to grab the order ID from the URL and use that to make a fetch request to the backend to get the full order info. This order info should give you everything you need on the `CheckoutComplete` page

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Done.

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Done. 
Subcomponent CheckoutMessageCard is implemented.

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Fixed.

AlonaVladymyrovaTrinity:
All mentor's recommendations have been successfully implemented.

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,120 @@
...
+              updateStatus={updatePaymentStatus}
+            />
+          </Grid>
+
+          <Grid item xs={12} sm={6}>
+            <Paper className={`${style.paper} ${style.rightPaper}`}>
+              {!isPaymentSuccessful && (
+                <Typography variant="h6">
+                  Payment was unsuccessful! Sorry for the inconvenience.
+                </Typography>
```
akosasante: Right now; I get a flash of the "payment was unsuccessful" message before the real result loads because when the component first loads and we haven't yet gotten the info from stripe, `isPaymentSuccessful` starts off as false. Maybe it would make sense to only show this if we've finished loading:

```suggestion
              {!loading && !isPaymentSuccessful && (
                <Typography variant="h6">
                  Payment was unsuccessful! Sorry for the inconvenience.
                </Typography>
```

src/components/Purchase/CheckoutComplete.js
```diff
@@ -0,0 +1,120 @@
...
+              )}
+
+              {isPaymentSuccessful && (
+                <>
+                  {isLoading && <Loader className="small-spinner" />}
```
akosasante: Maybe this `isLoading` check can be at top level? So if loading is still happening, then no matter the other variables, we'd show the loader. If loading is complete, then we can go down the list of figuring out which content to display

so we'd have something like the following which reduces the nesting and variables to check at each branch a bit

```
{isLoading && <Loader />}
{!isLoading && (
  <>
  {!isPaymentSuccessful && ...payment unsucessful content...}
  {errorMessage && ...error content...}
  {isPaymentSuccessful && ...success content...}
  </>
)}
```

akosasante (approved):
small suggestion to reduce nesting and a comment about brief flash of the unsuccessful state but otherwise everything looks good; I was able to test the order flow successfully :+1:

src/components/Purchase/CheckoutComplete.js
akosasante: https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/997602fe-e793-466b-bc80-9f542a350ac9

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: I made these changes. I hope I understood you correctly. Please review.

src/components/Purchase/CheckoutComplete.js
AlonaVladymyrovaTrinity: Yes. Good point! The glitch was indeed present.

akosasante: Sorry for the continued back and forth on this but I think I'm still seeing the glitch :thinking: 

It seems to be because we disable the `isLoading` state as soon as the Order API response has been returned even if the payment status from stripe hasn't been returned yet. One solution might be to have two separate loading states:

```javascript
// CheckoutComplete.js

  const [isLoading, setIsLoading] = useState(true);
  const [isPaymentLoading, setIsPaymentLoading] = useState(true);
  // this variable "derives state" from the above to state variables. 
  // It will evaluate to true when both the order and the  payment status have been loaded
  const bothLoaded = !isLoading && !isPaymentLoading;
  
  // Callback function to update isPaymentSuccessful
  const updatePaymentStatus = (isSuccess) => {
    setIsPaymentLoading(false);
    setIsPaymentSuccessful(isSuccess);
  };
  
  // in the return block show the loader if both responses haven't loaded yet, otherwise check for payment status or errors:
            <Grid item xs={12} sm={6}>
            <Paper className={`${style.paper} ${style.rightPaper}`}>
              {!bothLoaded && <Loader className="small-spinner" />}
              {bothLoaded && (
                <>
                ...etc
                </>
```

```javascript
// PaymentStatus.js

// Before, this was only ever either true or false. 
// But that meant that our parent component could not differentiate between "paymentsuccess is false because the payment failed" and "payment status is false because  we haven't loaded it yet and that's the default state"
// So we're essentially adding a third possible value for this state "null" to mean "payment status has not yet been retrieved"
// Then the parent components can rely on this to determine whether or not to set paymentSucces values yet
  const [isSuccess, setIsSuccess] = useState(null); 
```

```javascript
// CheckoutMessageCard.js

// Same idea as in PaymentStatus, start off the value as "null" to indicate "not yet defined/loaded"

const [isPaymentSuccessful, setIsPaymentSuccessful] = useState(null);

// only update the payment successful state here and in the parent component if we actually have gotten a non-null value from PaymentStatus component
  
  // Callback function to update isPaymentSuccessful
  const updatePaymentStatus = (isSuccess) => {
    if (isSuccess !== null) {
      setIsPaymentSuccessful(isSuccess);
    }
  };

  useEffect(() => {
    if (isPaymentSuccessful !== null) {
      updateStatus(isPaymentSuccessful);
    }
  }, [updateStatus, isPaymentSuccessful]);
```

AlonaVladymyrovaTrinity: Hi Akos, I think I found another alternative way to fix the loading glitch. I decided to take the initiative and offer my own solution. It seems to me that it has a little less lines of code, however, I can implement your approach if my solution does not look optimal enough. In my approach, instead of creating another loading state, I simply pass the value from isLoading through a prop to the child component. Yes, this is prop drilling, but on the other hand, it seems to me that in this way we add less lines of code. Your opinion on this matter is very interesting to me.

AlonaVladymyrovaTrinity: I reverted my changes and implemented the previously recommended approach.

akosasante (approved):
Looks great, no longer seeing the glitch. Really fantastic work on this one Alona!

:rocket:

AlonaVladymyrovaTrinity:
Thank you!!!

PR #32, opened 2023-08-28
Ps 57 create a redux store for handling the shopping cart state

AlonaVladymyrovaTrinity:
Pull Request Summary:
**This PR introduces shopping cart state management using Redux. The changes include:**

1. **Added `src/redux/store.js`:**
   - Created a Redux store using `configureStore`.
   - Configured a reducer for the shopping cart (`cart`) and potential other reducers.
   - Exported the configured store.

2. **Added `src/redux/shoppingCart.js`:**
   - Implemented a shopping cart reducer with `createSlice`.
   - Added functions to save/load cart state to/from `localStorage`.
   - Initialized the cart state from `localStorage` or defaults.
   - Implemented functions such as adding items, reducing items, and clearing the cart.

3. **Modified `src/index.js`:**
   - Imported `Provider` from `'react-redux'` for Redux store integration.
   - Provided the created Redux store to the app using `Provider`.
   - Ensured components within `App` can access the Redux store.

These comprehensive changes establish efficient shopping cart management through Redux.


src/redux/shoppingCart.js
```diff
@@ -0,0 +1,76 @@
...
+      const { album, quantity } = action.payload;
+      const albumId = album.id;
+
+      if (state.items[albumId]) {
+        const remainingQuantity = Math.max(
+          state.items[albumId].quantity - quantity,
+          0
+        );
+        state.totalQuantity -=
+          state.items[albumId].quantity - remainingQuantity;
```
akosasante, line 51: Hmm, if I'm reading this correctly I don't think it works in the following example case:

```javascript
{  
  items: [ {album: {id: "id1", ...etc}, quantity: 3 }],
  totalQuantity: 3
}
```

And a `reduceItem` event comes in with the payload:

```javascript
{ album: {id: "id1", ...etc}, quantity: 10 }
```

So `reamainingQuantity` will be 0. And then here on lines 50-51, we'd be doing 

`state.totalQuantity =`   
`state.totalQuantity - state.items["id1"].quantity - remainingQuantity =`   
`3 - (3 - 10) =`   
`-10`?

src/redux/shoppingCart.js
```diff
@@ -0,0 +1,76 @@
...
+        state.subtotal -=
+          album.price * (state.items[albumId].quantity - remainingQuantity);
```
akosasante, line 53: Same as above comment, let's double check if this works in the case where the payload is trying to decrement more than what we have in the cart.

akosasante (commented):
1 question but otherwise looks good

AlonaVladymyrovaTrinity:
Thanks for your recommendations. I carefully double-checked my code and tested it with the example you provided, as well as with other similar test examples. Before doing the math you mentioned above, I do this check in my code (line 45-49) 
```
     if (state.items[albumId]) {
        const remainingQuantity = Math.max(
          state.items[albumId].quantity - quantity,
          0
        );
```
This way I make sure values don't go into negatives! Please double check my Pull Request again. Thank you.
Screenshot 2023-08-28 at 10 25 55 PM
Screenshot 2023-08-28 at 10 26 20 PM

akosasante (approved):
Ah yep I tested the way you've shown and it looks like it's all working well; sorry about my bad mental math lol :sweat\_smile:

PR #33, opened 2023-08-28
PS-62 Create a react component for the add to cart button

AlonaVladymyrovaTrinity:
Pull Request Summary:
**For the task "Create a React component for the 'Add to Cart' button," the following changes have been made:**

**Modified src/App.js:**

- Added an import for the ShoppingCart component. (temporarily for review and testing)
- Added a route for /shoppingCart that renders the ShoppingCart component. (temporarily for review and testing)
- Url for testing http://localhost:3000/shoppingCart

**Modified src/components/Album.js:**

- Imported the newly created AddToCartButton component.
- Replaced the previous "Buy" button with the AddToCartButton component.

**Added src/components/ShoppingCartComponents.js:** (temporarily for review and testing)

- Created a simple ShoppingCart component for testing purposes.
- Displays items in the cart, along with subtotal, total quantity, and total amount.
- Provides a "Clear Cart" button that dispatches the clearCart action.

**Added src/components/layout/AddToCartButton/AddToCartButton.js:**

- Implemented the AddToCartButton component.
- Handles adding/removing albums to/from the cart based on user actions.
- Displays different labels based on whether an album is already in the cart or not.
- Handles user authentication status and redirects to sign-in if not logged in.


src/components/layout/AddToCartButton/AddToCartButton.js
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  return (
+    <Button
+      variant="contained"
+      color="primary"
+      className="buy-button"
+      aria-label={itemsInCart[album.id] ? 'Remove from cart' : 'Add to Cart'}
+      onClick={handleAddToCart}
+      disabled={!isUserLoggedIn}
```
akosasante: since this is disabled when I'm not logged in; I can't actually click it to get redirected to `signIn`. A couple options I can think of here:

*   don't disable it; and just let it redirect users when they click it. only downside here is it can be a bit surprising for the user when they click that and it immediately sends them away from this page. But otherwise I think it's okay and we can tweak the ux for that later. 
*   disable but show a popover/tooltip when they hover over the button that directs them to sign in. Some notes for this approach, you need to wrap the Button element in a span according to the Material-UI docs for the Tooltip component to work on disabled elements: [https://mui.com/material-ui/react-tooltip/#disabled-elements.](https://mui.com/material-ui/react-tooltip/#disabled-elements.) As well, if you want to use HTML inside the `Tooltip` title prop, you'll need to use a React Fragment for the title instead of a plain string. When the button is not disabled (user is logged in) you can set the Tooltip's title to an empty string and then it won't show up

src/components/Album.js
```diff
@@ -10,9 +11,11 @@ function Album({ album }) {
         <p className="album-title">{album.albumName}</p>
         <p className="album-artist">{album.artistName}</p>
         <div className="button-container">
-          <Button variant="contained" color="primary" className="buy-button">
+          {/* <Button variant="contained" color="primary" className="buy-button">
             Buy
-          </Button>
+          </Button> */}
+          {/*AddToCartButton component */}
+          <AddToCartButton album={album} />{' '}
```
akosasante: Is the `{' '}` here for spacing purposes or something else?

akosasante (commented):
Looks great and thank you for the clear and detailed PR description with testing steps! :heart\_eyes:

src/components/Album.js
AlonaVladymyrovaTrinity: It's Prettier's correction.  I removed it in my last commit.

src/components/layout/AddToCartButton/AddToCartButton.js
AlonaVladymyrovaTrinity: Good point! Thanks for providing a few options how to solve this problem. I chose the first option. The changes were made with the last commit.

src/components/Album.js
akosasante: ah okay if the formatter wants to put it there then that's fine, we'll live with it :sweat\_smile:

akosasante (approved):
LGTM! :rocket:
(just make sure linting and prettier formatting is happy before merging)

PR #35, opened 2023-08-30
PS-108 Create a react component for the wishlist button

AlonaVladymyrovaTrinity:
**Pull Request Description:**

**Title:** Implementing Add to Wishlist Button Component

**Description:**
This pull request introduces the "AddToWishlistButton" component to the application. This component provides the functionality for users to add or remove albums to/from their wishlists. The button's behavior changes based on the user's login status and whether the album has already been added to the wishlist.

**Changes Made:**
- Added a new component named "AddToWishlistButton" responsible for managing the addition and removal of albums from the wishlist.
- The "AddToWishlistButton" component provides a seamless way for users to interact with their wishlists, enhancing their experience by allowing them to manage their favorite albums effortlessly.
- Users can seamlessly add or remove albums from their wishlist using the same button. The button intelligently switches between "Add to Wishlist" and "Remove from Wishlist" based on the album's current wishlist status, streamlining the interaction process.
- Users are directed to the sign-in page if they are not logged in and attempt to add an album to the wishlist.

**Related Back-end Pull Request**
https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/42

**Testing steps**

1. Click the button to either "Add to Wishlist" or "Remove from Wishlist."
2. After each button click, examine the wishlist collection in the database.
3. Verify that the album's status in the wishlist corresponds to the action performed.
4. Validate that the database accurately reflects changes based on the button's toggle functionality.
5. Click the button while logged out and observe whether you're redirected to the login page.

**Screenshots:**
Screenshot 2023-08-30 at 12 54 29 PM
Screenshot 2023-08-30 at 1 04 05 PM


src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -0,0 +1,111 @@
...
+import axios from 'axios';
+import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
+import { useNavigate } from 'react-router-dom';
+
+const AddToWishlistButton = ({ album }) => {
+  const [isAdded, setIsAdded] = useState(false);
+  const { status } = useAuth();
+  const navigate = useNavigate();
+  const isUserLoggedIn = status === AuthStatus.LoggedIn;
+
```
akosasante, line 14: Generally for all these `axios` calls; let's opt to use the base instance/client that is set up here: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/PS-108-Create-a-React-component-for-the-wishlist-button/src/apis/axiosClient.js](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/PS-108-Create-a-React-component-for-the-wishlist-button/src/apis/axiosClient.js)

That way it already does `withCredentials` for us. And in the future if we want to make any other changes that should effect _all_ http calls from the frontend, there's a single place to make that change, insetad of having to edit every component.

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -0,0 +1,111 @@
...
+  const fetchOrCreateWishlist = async () => {
+    try {
+      const response = await axios.post(
+        'http://localhost:8000/api/v1/wishlist',
+        { albumId: album._id },
```
akosasante: I'm reviewing Yelena's ticket that adds this endpoint and discussing if we need to pass in an album ID. Since the frontend is already going to make a subsequent call to add items to the wishlist via the `PATCH` endpoint, I'm not sure if it makes sense to accept an album ID on the create wishlist endpoint.

One thing I think we should do (time permiting) is have the backend create and return an empty list when the user is first created/registered or attach their existing wish list when they register. The app can then store that in redux/localstorage and use it throughout without having to necessarily make this fetchOrCreateWishlist call from within every `Add To Wishlist` button that appears on the page. 

For now though, let's keep this call, just without passing in the album id since the backend is going to ignore that anyway.

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
AlonaVladymyrovaTrinity: Done.

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
AlonaVladymyrovaTrinity: Done.

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -0,0 +1,100 @@
+import React, { useState, useEffect } from 'react';
+import Button from '@mui/material/Button';
+import axiosInstance from '../../../apis/axiosClient'; // axios instance
+import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
+import { useNavigate } from 'react-router-dom';
+
+const apiBaseURL = process.env.REACT_APP_API_BASE_PATH;
+const wishlistEndpoint = `${apiBaseURL}/wishlist`;
```
akosasante: The axios base instance already takes care of setting up the `REACT_APP_API_BASE_PATH` part (locahost:8000/api/v1 or musicstore.onrender.com/api/v1), so all we need to add here in this file is just the "wishlist" part:

```suggestion
const apiBaseURL = process.env.REACT_APP_API_BASE_PATH;
const wishlistEndpoint = '/wishlist';
```

akosasante (approved):
One tiny thing to adjust but then feel free to merge if all CI is passing :+1:

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
AlonaVladymyrovaTrinity: And after that I will need to make changes to my API request?  Here:
`${wishlistEndpoint}/${wishlistId}/add_album/${album._id}`

AlonaVladymyrovaTrinity: Or I should just remove this line:
```
- const apiBaseURL = process.env.REACT_APP_API_BASE_PATH;
+ const wishlistEndpoint = '/wishlist';
```

akosasante: Nope; it should still work without needing to change those. Because before; you were setting `wishlistEndpoint` to be `process.env.REACT_APP_API_BASE_PATH` which equaled: `http://localhost:8000/api/v1` on local. Because the address started with the scheme (http) it _**overrides**_ the one that was set in the base axios instance (even though in this case, it's overriding it with basically the same base + `/wishlist`. 

But if we just set `wishlistEndpoint` to only be `/wishlist`, then axios will see that that is not a full URL host; so it will check the instance being used to see if we've set a base URL there (we have [here](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/src/apis/axiosClient.js#L5)) and it will instead just append `/wishlist` to that. So the end result is the same, but with a smidge less lines of code to be repeated here in this file.

akosasante: > Or I should just remove this line:
> 
> ```
> - const apiBaseURL = process.env.REACT_APP_API_BASE_PATH;
> + const wishlistEndpoint = '/wishlist';
> ```

Yes we can remove `const apiBaseURL = process.env.REACT_APP_API_BASE_PATH;`
And we need just `const wishlistEndpoint = '/wishlist';`

AlonaVladymyrovaTrinity: Done!

PR #36, opened 2023-08-31
clean album design

dayofthetech:
I was able to change some code and introduced tooltip from mui to design a better album layout - I still need to find out if the css file is needed and if I can move the design into another component


src/components/AlbumGrid.css
akosasante, line 1: Yeah I think you're right it doesn't look like it's being imported anywhere which means it's not being used and everything still looks fine :+1: So this file can be deleted for now.

akosasante (approved):
Looks great!

PR #39, opened 2023-09-07
album details page and responsive design

Musab-Kara:
PS-77,PS-78,PS-79,PS-80 ( For album details page)
PS-173,PS-174,PS-175,PS-177 (For responsive details)

=> integrated API to get reviews of the particular albums and displayed them on the dialog box on the album detail page.
=> integrated API for adding new reviews for particular albums on the album detail page.
=> integrated API for deleting reviews for particular albums on the album detail page.
=> integrated API for edit reviews for the particular albums on the album detail page.
=> set a condition for allowing to add a review if the user has not written one for that album on the album detail page.
=> set conditions for allowing delete and edit only those reviews which review added by the the login user on the album detail page.
=> set a responsive design in the navbar page on the music app project.
=> set responsive design in the sign-up and sign-in page on the music app page.


.env
akosasante, line 1: We don't want to delete this `.env` file since it is needed so that the environment variables for the Base URL for api calls are retained for all branches

src/components/Navbar.js
```diff
@@ -13,29 +32,91 @@ const Navbar = () => {
...
+            </>
+          )}
+        </Box>
+        <Box sx={{ display: { xs: 'block', sm: 'none' } }}>
+          <IconButton
+            size="large"
+            aria-label="account of current user"
+            aria-controls="menu-appbar"
+            aria-haspopup="true"
+            onClick={() => setAnchorElUser(true)}
```
akosasante: I'm reading the material-ui documentation and it looks like `Menu` component expects `anchorEl` to be an HTML Element or a function that returns an HTML element: [https://mui.com/material-ui/api/menu/#Menu-prop-anchorEl](https://mui.com/material-ui/api/menu/#Menu-prop-anchorEl)

But here you're setting it to a boolean, which is resulting in some errors that say `"expected one of type [function]"`:



If we look at the example for positioned menu here in the docs: [https://mui.com/material-ui/react-menu/#positioned-menu](https://mui.com/material-ui/react-menu/#positioned-menu)

We can see that they use the `onClick` handler and set the "anchorEl" value to the button that was clicked. This tells the `Menu` component where to anchor/position the menu that is displayed:

```suggestion
            onClick={event => setAnchorElUser(event.currentTarget)}
```

src/components/Navbar.js
```diff
@@ -13,29 +32,91 @@ const Navbar = () => {
...
+                <MenuItem onClick={handleCloseUserMenu}>
+                  <Button color="inherit" component={Link} to="/profile">
+                    Profile
+                  </Button>
+                </MenuItem>
+                <MenuItem onClick={handleCloseUserMenu}>
+                  <Logout />
+                </MenuItem>
+              </>
+            )}
```
akosasante: I'm seeing the following error in the dev console:



The `<> ... </>` tags are shorthand for `React.Fragment`. The error is saying that `Menu` component doesn't allow having those as children; so we need to remove them here while still being able to pass in multiple children. The error message says "consider providing an array instead" so let's try to do exactly that:

```suggestion
            {!isLoggedIn && ([
              <MenuItem key='login' onClick={handleCloseUserMenu}>
                <Button color="inherit" component={Link} to="/signIn">
                  Login
                </Button>
              </MenuItem>,
              <MenuItem key='signup' onClick={handleCloseUserMenu}>
                <Button color="inherit" component={Link} to="/register">
                  Signup
                </Button>
              </MenuItem>
              ]
            )}
            {isLoggedIn && ([
              <MenuItem key='profile' onClick={handleCloseUserMenu}>
                <Button color="inherit" component={Link} to="/profile">
                  Profile
                </Button>
              </MenuItem>,
              <MenuItem key='logout' onClick={handleCloseUserMenu}>
                <Logout />
              </MenuItem>
              ]
            )}
```

src/components/Album.js
akosasante, line 1: We're adding a lot of new code to this component that is specific to adding/removing/editing reviews. Could those be extracted into a child component instead?

Ending up with a file structure something like:

```
components/
├─ album/
│  ├─ Album.js
│  ├─ AlbumReviews.js
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                      </Box>
+                    ) : (
+                      <>
+                        <Typography className="rate-data">
+                          {val.title}
+                        </Typography>
+                        <Typography>Rate: {val.rating}</Typography>
+                        <Typography>
+                          Posted on{' '}
+                          {new Date(val.createdAt).toLocaleDateString()}
```
akosasante: It would be nice to also include "Posted by: \<user name>". That may require an update on the backend to populate that data (you can feel free to do it or ask backend team for help getting that change in if needed)

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+      title: '',
+      comment: '',
+      rating: '',
+    });
+  };
+
+  const show = () => {
+    const data = reviewList.filter((val) => val.user === loginID);
+    return data.length === 0 && isLoggedIn;
+  };
```
akosasante: This is fine as-is but here is a suggestion for making a it a one-line variable rather than a function:

```suggestion
  const show = isLoggedIn && !reviewList.some((val) => val.user === loginID);
```

  
[Array.some ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) will return true if the boolean operation inside returns true for at least one element in the list. We want to know if _none_ of the elements in the list belong to the user so we negate with `!`

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+      console.error('Error update review:', error);
+    }
   };
 
+  useEffect(() => {
+    if(isLoggedIn){
+      const ID = JSON.parse(localStorage.getItem('auth_user'));
+      ID.userId ? setLoginID(ID.userId) : setLoginID(ID.user.id);
+    }
+  }, []);
```
akosasante: Instead of this; we can also just use the `useAuth` hook to get the user as it will also store that data:

```suggestion
  const { user } = useAuth
  const loginID = user?.user?.id
```

In which case we don't need to have a `useState` variable for loginID since it can just be derived from the value of that hook so line 46 can be deleted

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                          id="outlined-basic"
+                          label="Enter rating between 1 to 5"
+                          value={editObj.rating}
+                          onChange={(e) =>
+                            setEditObj({ ...editObj, rating: e.target.value })
+                          }
+                          min={1}
+                          max={5}
+                          variant="outlined"
+                        />
```
akosasante: Looking at the examples here: [https://mui.com/material-ui/react-text-field/#form-props](https://mui.com/material-ui/react-text-field/#form-props)

you need to set the `type` of the field to tell it to only accept numbers.

And to pass in min and max we need to use `InputProps`: [https://stackoverflow.com/questions/47798104/set-min-max-on-textfield-type-number](https://stackoverflow.com/questions/47798104/set-min-max-on-textfield-type-number)

```suggestion
                        <TextField
                          id="outlined-basic"
                          type="number"
                          label="Enter rating between 1 to 5"
                          value={editObj.rating}
                          onChange={(e) =>
                            setEditObj({ ...editObj, rating: e.target.value })
                          }
                          InputProps={{ inputProps: { min: 1, max: 5 } }}
                          variant="outlined"
                        />
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+              alignItems: 'center',
+            }}
+          >
+            {album.albumName}
+            <Box sx={{ cursor: 'pointer' }} onClick={handleCloseDialog}>
+              <CloseIcon />
+            </Box>
+          </DialogTitle>
           <DialogContent>
             <DialogContentText>
```
akosasante: I'm seeing a bunch of these errors:



This is occuring because `<DialogContentText>` ends up being rendered as a `<p>` tag; and proper HTML does not allow nesting block tags inside a `<p>`. We can avoid this by using a regular `<div>` here instead:

```suggestion
            <div>
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                    </Box>
+                    <Box>Average Rating: {album.averageRating}</Box>
+                  </Box>
+                </Box>
+
+                <Typography className="title" variant="h5">
+                  Review:
+                </Typography>
+                {reviewList.map((val) => (
+                  <div className="rate-card">
```
akosasante: Anytime we iterate over a list in React we have to make sure the first child has the `key` attribute. This allows React to better know how to make render updates when one of the items in the list has a state change. [https://react.dev/learn/rendering-lists](https://react.dev/learn/rendering-lists)

```suggestion
                {reviewList.map((val) => (
                  <div className="rate-card" key={val._id}>
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                        id="outlined-basic"
+                        label="Enter rating between 1 to 5"
+                        value={formData.rating}
+                        onChange={(e) =>
+                          setFormData({ ...formData, rating: e.target.value })
+                        }
+                        min={1}
+                        max={5}
+                        variant="outlined"
+                      />
```
akosasante:
```suggestion
                      <TextField
                        id="outlined-basic"
                        type="number"
                        label="Enter rating between 1 to 5"
                        value={formData.rating}
                        onChange={(e) =>
                          setFormData({ ...formData, rating: e.target.value })
                        }
                        InputProps={{ inputProps: { min: 1, max: 5 } }}
                        variant="outlined"
                      />
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                    <Box>
+                      Release Date:{' '}
+                      {new Date(album.releaseDate).toLocaleDateString()}
+                    </Box>
+                    <Box>Average Rating: {album.averageRating}</Box>
+                  </Box>
+                </Box>
+
+                <Typography className="title" variant="h5">
+                  Review:
```
akosasante:
```suggestion
                  Reviews:
```

src/components/Album.js
```diff
@@ -30,82 +35,332 @@ import AddToWishlistButton from './layout/AddToWishlistButton/AddToWishlistButto
...
+                            Save
+                          </Button>
+                        </Box>
+                      </Box>
+                    ) : (
+                      <>
+                        <Typography className="rate-data">
+                          {val.title}
+                        </Typography>
+                        <Typography>Rate: {val.rating}</Typography>
```
akosasante:
```suggestion
                        <Typography>Rating: {val.rating}</Typography>
```

akosasante (commented):
Looking great. Some extra suggestions (could be done in a separate PR or just left out completely it's up to you):

*   use Snackbar component to indicate to user when something goes wrong with saving their review: [https://mui.com/material-ui/react-snackbar/](https://mui.com/material-ui/react-snackbar/)
*   use Rating component to show review rating as stars: https://mui.com/material-ui/react-rating/

src/components/Album.js
Musab-Kara: considering there is a demo on Monday, I will defer this for now and will take care of it in the future

akosasante (approved):
LGTM :ship:

akosasante:
You'll want to do a `npm run format` to reformat code so that the Github CI passes @Musab-Kara

PR #40, opened 2023-09-07
Ps 81 create sidebar for order details

JafariM:


package-lock.json
```diff
@@ -24,6 +24,7 @@
         "react-dom": "^18.2.0",
         "react-payment-inputs": "^1.1.9",
         "react-redux": "^8.1.2",
+        "react-redux": "^8.1.2",
```
akosasante, line 29: Hmmt, it seems a bit odd that this `react-redux` line shows up here twice despite no changes to `package.json`. I'd like to ask that you delete the `package-lock.json` file and then reinstall by doing `npm install`. If it still shows up twice then I guess it's expected

src/components/OrderSidebar.js
```diff
@@ -0,0 +1,104 @@
...
+        border: '1px solid grey',
+        padding: 2,
+      }}
+    >
+      <List>
+        <Typography variant="h6">Order Summery</Typography>
+        {Object.values(itemsInCart.items).map((item) => {
+          return (
+            <>
+              <ListItem key={item.album.id}>
```
akosasante: I'm seeing the following erorr in the dev console:



this is happening because anytime we iterate over an array (eg: with `.map(...)`) in React, we have to give the results of that iteration a `key` so that React knows how to handle any updates to that list: [https://react.dev/learn/rendering-lists](https://react.dev/learn/rendering-lists)

In this case, you've put the key on `ListItem` but that's not actually the main child element returned by the `map(...)` here, rather the main child is `<>` , which is shorthand for `<React.Fragment>` (or if you `import { Fragment} from 'react'` then you can do just `<Fragment>`) . But you can't put the `key` on the shorthand version so you have to use one of the longer forms: [https://react.dev/reference/react/Fragment#:~:text=If%20you%20want%20to%20pass%20key%20to%20a%20Fragment%2C%20you%20can%E2%80%99t%20use%20the%20%3C%3E...%3C/%3E%20syntax.%20You%20have%20to%20explicitly%20import%20Fragment%20from%20%27react%27%20and%20render%20%3CFragment%20key%3D%7ByourKey%7D%3E...%3C/Fragment%3E.](https://react.dev/reference/react/Fragment#:~:text=If%20you%20want%20to%20pass%20key%20to%20a%20Fragment%2C%20you%20can%E2%80%99t%20use%20the%20%3C%3E...%3C/%3E%20syntax.%20You%20have%20to%20explicitly%20import%20Fragment%20from%20%27react%27%20and%20render%20%3CFragment%20key%3D%7ByourKey%7D%3E...%3C/Fragment%3E.)

```suggestion
        {Object.values(itemsInCart.items).map((item) => {
          return (
            <React.Fragment key={item.album.id}>
              <ListItem>
```

src/components/OrderSidebar.js
```diff
@@ -0,0 +1,104 @@
...
+  Button,
+  Typography,
+  Box,
+} from '@mui/material';
+import { useNavigate } from 'react-router-dom';
+import { clearCart } from '../redux/shoppingCart';
+
+const OrderSidebar = () => {
+  //path to env variable
+  const envPath = process.env.REACT_APP_API_BASE_PATH;
```
akosasante: no longer needed

src/components/Purchase/CheckoutPage.js
```diff
@@ -21,34 +22,40 @@ const total = 140.0;
...
-  }, [subtotal, album, quantity, tax, total, setClientSecret]);
+          setClientSecret(response.data.clientSecret);
+        } catch (error) {
+          // Handle error here
+          console.error('Error fetching data:', error);
+        }
+      };
+      fetchData();
+    }
+  }, [setClientSecret]);
```
akosasante: This main external dependency of this useEffect is now `location.state`. We want it so that this useEffect hook only runs when that value changes/is updated:

```suggestion
  }, [location.state]);
```

src/components/OrderSidebar.js
```diff
@@ -0,0 +1,104 @@
...
+
+  const navigate = useNavigate();
+  const dispatch = useDispatch();
+
+  // values come from Redux store
+  const itemsInCart = useSelector((state) => state.cart);
+  const subtotal = useSelector((state) => state.cart.subtotal); //total price of all items
+  const tax = useSelector((state) => state.cart.tax);
```
akosasante, line 23: The tax in the cart is the percentage (eg: 0.15 for 15%). But right now we're displaying it as a dollar amount:



If we want to display the dollar amount of how much will be added to the subtotal due to taxes, then we should multiply it by `subtotal`:

```suggestion
  const tax = useSelector((state) => state.cart.tax) * subtotal;
```

src/components/Purchase/CheckoutPage.js
```diff
@@ -21,34 +22,40 @@ const total = 140.0;
...
+              orderItems: orderData.orderItems, // Use order data received from state
+              subtotal: orderData.subtotal,
+              tax: orderData.tax,
+              total: orderData.total,
+            },
+            {
+              withCredentials: true,
+              headers: { 'Content-Type': 'application/json' },
+            }
+          );
```
akosasante: Let's import the base axios client from the apis folder so that we don't have to handle `REACT_APP_API_BASE_PATH`, `withCredentials`, and `headers` all manually here

src/components/Purchase/CheckoutPage.js
```diff
@@ -21,34 +22,40 @@ const total = 140.0;
...
+    const { orderData } = location.state || {};
 
-        setClientSecret(response.data.clientSecret);
-      } catch (error) {
-        // Handle error here
-        console.error('Error fetching data:', error);
-      }
-    };
+    if (orderData) {
+      setOrderData(orderData);
```
akosasante: no longer needed, since we're receiving the order data from the router state

src/components/Purchase/CheckoutPage.js
```diff
@@ -21,34 +22,40 @@ const total = 140.0;
 
 const CheckoutPage = (/*{ album, quantity, subtotal, tax, total}*/) => {
   const [clientSecret, setClientSecret] = useState('');
-
+  const [orderData, setOrderData] = useState(null);
```
akosasante: Not needed as comonent state, since we only use the orderData inside useEffect

src/components/Purchase/CheckoutPage.js
```diff
@@ -21,34 +22,40 @@ const total = 140.0;
...
+      // Create PaymentIntent as soon as the page loads
+      const fetchData = async () => {
+        try {
+          const response = await axios.post(
+            `${process.env.REACT_APP_API_BASE_PATH}/orders`,
+            {
+              orderItems: orderData.orderItems, // Use order data received from state
+              subtotal: orderData.subtotal,
+              tax: orderData.tax,
+              total: orderData.total,
```
akosasante: I'm finding that the backend will return a 400 due to Stripe not expecting extra decimals:





I've put in a quick fix to this in the `main` branch of the backend: Code-the-Dream-School/dd-prac-team4-back@dd82ab4#diff-76a253026a64d07d6b9b82dd96a6d3b4820ed874c4c25e8d0bf4f5b15cbd19e6R30

src/components/Purchase/CheckoutPage.js
akosasante, line 1: The values on line 15-21 were just temporary until this ticket was ready; so we can remove them now; and instead just pass the total to the `CheckoutForm` component

akosasante: That's where the hardcoded "Payment Amount $100.00" is coming from

src/components/Purchase/CheckoutPage.js
akosasante: Instead, I would recommend moving te `location.state` extraction out of useEffect and putting here:

```suggestion
  // Extract order data from navigation.state
  const { orderData } = location.state || {};
```

that way, it'll be easier to pass the order data.total value into the CheckoutForm component

src/components/Purchase/CheckoutPage.js
akosasante: It might still be helpful to only pass in numbers with exactly 2 decimal digits. A function we can use to help do this: https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary

```javascript
const roundedNumber = (num) => Math.round((num + Number.EPSILON) * 100) / 100
```

Then here we can round all the dollar values (subtotal and total) when passing it to the POST /orders request; and when passing the **total** into the CheckoutForm on line 74 (instead of subtotal)

src/components/ShoppingCart.js
akosasante, line 1: Is this file being used somewhere; or was it just for testing? Will it be removed before merging?

src/components/WishListView.js
akosasante, line 1: Is this file being used somewhere/expected to be part of the changes for the order sidebar?

package-lock.json
JafariM: I re install package-lock.json and it is only one now

src/components/OrderSidebar.js
JafariM: Done

src/components/OrderSidebar.js
JafariM: Removed

src/components/Purchase/CheckoutPage.js
JafariM: Updated

src/components/ShoppingCart.js
JafariM: Alona created it for testing, I will remove it

src/components/Purchase/CheckoutMessageCard.js
```diff
@@ -59,7 +59,7 @@ const CheckoutMessageCard = ({
                     totalItemsPurchased === 1 ? 'album' : 'albums'
                   } has been completed.`}
                   <br />
-                  {`A confirmation email was sent to ${userEmail}`}
+                  {/* {`A confirmation email was sent to ${userEmail}`} */}
```
akosasante: I think how this was before was correct; 

`user` from `useAuth` returns a nested object like:



so we do have to access the email by going `user.user.email`

akosasante (approved):
One small note but otherwise looks good to me after merging latest changes from main and fixing any conflicts + getting CI passing

akosasante:
👍  let me know if you need any assistance with resolving the merge conflicts here Masooma

PR #42, opened 2023-09-08
Fetch user information for user profile page

Rjjayy:


src/index.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
-import ReactDOM from 'react-dom/client';
+import ReactDOM from 'react-dom';
+// import { Button } from 'reactstrap';
```
akosasante: this can be removed :)

src/index.js
```diff
@@ -1,10 +1,14 @@
...
 import App from './App';
 import { Provider } from 'react-redux';
 import store from './redux/store';
 import { BrowserRouter } from 'react-router-dom';
 import reportWebVitals from './reportWebVitals';
+import UserProfile from './components/UserProfile'; // Import the UserProfile component
+// import 'bootstrap/dist/css/bootstrap.min.css';
```
akosasante: this bootstrap comment can be removed :)

src/components/getUserProfile.js
akosasante, line 1: If this file is all commented out we can delete it now

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+  getImageListItemBarUtilityClass,
+} from '@mui/material';
+import { useNavigate } from 'react-router-dom';
+// import { getUserProfile } from './api';
+import { Avatar } from '@mui/material';
+import axios from 'axios';
+
+export default function PersonalProfile() {
+  const navigate = useNavigate();
+  const { user, status } = useAuth();
```
akosasante: In order to use `useAuth` you must import it: `import { useAuth } from '@akosasante/react-auth-context';`

akosasante: The `useAuth` hook returns a `user` object that looks like the following:

```json
{
  "user": {
    "_id": "64ef5e90086e1c19b888c72c",
    "name": "Akos",
    "username": "akos",
    "email": "akos123456@example.com",
    "role": "admin",
    "__v": 0,
    "passwordResetExpiresOn": "2023-09-07T17:40:39.477Z",
    "passwordResetToken": "0b38692bf452eb218dd2cc65bf2c8a849bea99e7",
    "id": "64ef5e90086e1c19b888c72c"
  }
}
```

that already has the info we need for the profile and is stored when the user first logs in.

We can store the object in an `authUser` variable to use later in this file:

```suggestion
  const { user: authUser } = useAuth();
```

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+  const [userData, setUserData] = useState(null);
```
akosasante: Instead of setting the state from useEffect, we can just use the user that was returned from the `useAuth` hook in the previous line:

```suggestion
  const userData = authUser.user;
```

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+      .then((response) => {
+          setUserData(response.data); // Update user data in state
+        })
+        .catch((error) => {
+          // Handle errors, e.g., display an error message
+          console.error('Error fetching user data:', error);
+        });
+    }
+  }, 
+  [status]);
```
akosasante: Since the `useAuth` hook already includes the user info, we don't need to make a second call to the backend here. But for future reference, there's a few things to point out:

*   instead of using axios directly, we've been importing a base axios instance from the `apis` folder here: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/FetchUserProfilegit/src/apis/axiosClient.js](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/FetchUserProfilegit/src/apis/axiosClient.js)
    *   The reason we've been doing this is because we've set some default values there such as the Base URL, so that the code knows to make the backend request to `localhost:8000` when running in development but instead to [`https://musicstore.onrender.com`](https://musicstore.onrender.com) when running in production. It walso handles setting up cookies and any other headers that we may need
*   The backend API docs are all available on [https://musicstore.onrender.com/api-docs/](https://musicstore.onrender.com/api-docs/) . You'll see there that there is no endpoint `http://localhost:8000/api/v1/user` which is likely why the API call was failing when you were testing. Rather it's the plural form: `http://localhost:8000/api/v1/users` but that would give you _all users_ and is only available for admins. Instead, in this case we want to get info about the _current user_ so we woulduse `http://localhost:8000/api/v1/users/showMe` . 

But since in this case we don't actually need to make this API call, you can delete the whole `useEffect` block altogether.

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+    <TableCell>{userData?.name}</TableCell>
+  </TableRow>
+  <TableRow>
+    <TableCell>EMAIL</TableCell>
+    <TableCell>{userData?.email}</TableCell>
+  </TableRow>
+  <TableRow>
+    <TableCell>Password</TableCell>
+    <TableCell>{userData?.password}</TableCell>
+  </TableRow>
```
akosasante: Generally, it's not secure to show a user's password in plain text on a website. We should delete this row from the table

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+            className="img-fluid rounded-circle"
+            sx={{
+              width: '200px',
+              height: '200px',
+              maxWidth: '200px',
+              maxHeight: '200px',
+            }}
+          />
+        </div>
+        <Table responsive striped hover className="text-center mt-5">
```
akosasante: There are errors in the dev console about the `responsive`, `striped`, and `hover` attributes that you have here. 

I think you're thinking of bootstrap where those attributes are available. But in Material-UI they are not.



```suggestion
        <Table className="text-center mt-5">
```

src/components/UserProfile.js
```diff
@@ -0,0 +1,123 @@
...
+        <tbody>
```
akosasante: Since we're using material-ui components for the rest of the table we should here as well

```suggestion
        <TableBody>
```

(you'll need to make sure you also update the closing tag, and import the variable `TableBody` from mui)

src/index.js
```diff
@@ -1,15 +1,11 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
```
akosasante: ReactDOM is needed for line 36 to work otherwise the app won't load and you will see an error in the devconsole saying `ReactDOM is not defined`

akosasante (commented):
Looking and working great! Just make sure to do `git pull origin main --no-rebase` and fix any merge conflicts, then push and with that you should be good to go!

Or if you prefer you can fix merge conflicts within Github by clicking the "Resolve Conflicts" button. The only one I would not recommend that for is the `package-lock.json` file. For that one you'll want to follow the steps here: https://codethedreamteam4-2023.atlassian.net/wiki/spaces/PT4/pages/4194310/Resolving+package-lock.json+conflicts
Screenshot 2023-09-08 at 11 56 45 PM


Remaining work after merging in latest changes from main:
- **don't** delete `ReactDOM` from `index.js`
- delete `getUserProfile.js` file since it's not needed

akosasante (approved):

PR #43, opened 2023-09-08
Ps 65 create wish list view page

JafariM:


src/components/Album.js
akosasante, line 1: You'll want to pull latest changes from `main` . There was a previous merge that accidentally included some double code due to a mixup with resolving merge conflicts. Once you do that there should be just one `AlbumItemWrapper` element that wraps the code and has a single addtocart and addtowishlist buttons

src/components/AlbumsList.js
```diff
@@ -56,6 +58,29 @@ const AlbumsList = () => {
...
+        const wishlistData = response.data.wishlist;
+        setWishListId(wishlistData._id);
+        setWishListAlbums(wishlistData.albums);
+        // Store each album id from wishlist in local storage
+        wishlistData.albums.forEach((album) => {
+          localStorage.setItem(
+            `wishListAlbums-${album._id}`,
+            JSON.stringify(true)
+          );
+        });
```
akosasante: This is fine (and I know initially this is the same pattern the `addToWishlistButton` was using) but eventually if wishlist is large, we'll end up with a lot of entries in localstorage which may get difficult to manage or even cause browser slowdowns.

Let's try to store it in localstorage as an object:

```suggestion
        // Store object in local storage where keys are album id and value is the whole album
        const wishlistAlbumsToStore = wishlistData.albums.reduce((acc, album) => {
          acc[album._id] = album;
          return acc;
        }, {});
        localStorage.setItem("wishlistAlbums", JSON.stringify(wishlistAlbumsToStore));
```

or if you prefer to do it iteratively  (not with `Array.reduce`):

```suggestion
        // Store object in local storage where keys are album id and value is the whole album
        const wishlistAlbumsToStore = {};
        wishlistData.albums.forEach((album) => {
          wishlistAlbumsToStore[album._id] = album
        });
        localStorage.setItem("wishlistAlbums", JSON.stringify(wishlistAlbumsToStore));
```

src/components/wishList/WishListView.js
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useEffect } from 'react';
+import AlbumGrid from '../AlbumGrid';
+import axiosInstance from '../../apis/axiosClient';
+
+const WishListView = () => {
+  const [albums, setAlbums] = useState();
```
akosasante: If the albums are not yet loaded (we're waiting on the response from the API request), then right now `albums` will have a value of `undefined`.

This will cause the following error in `AlbumGrid` component when it tries to map over what it expects to be a list of albums



We can resolve this by setting the initial value of `albums` to an empty list; so when it first loads, AlbumGrid just gets an empty list, which it can try to call `.map(...)` on successfully but will just result in an empty grid:

```suggestion
  const [albums, setAlbums] = useState([]);
```

src/components/AlbumsList.js
```diff
@@ -26,6 +26,8 @@ const AlbumsList = () => {
   const [searchTerm, setSearchTerm] = useState(''); // for search input filed value
   const [message, setMessage] = useState(''); // for not found message
   const [errorMessage, setErrorMessage] = useState(''); // for error message
+  const [wishListAlbums, setWishListAlbums] = useState();
```
akosasante: Do we need `wishlistAlbums` to be in the component state? I'm not seeing this value be used elsewhere in this component, other than being set from within `useEffect`

src/components/wishList/WishListView.js
```diff
@@ -0,0 +1,35 @@
...
+        const wishlistData = response.data.wishlist;
+        setWishListId(wishlistData._id); //set the state for wishlist id
+        setAlbums(wishlistData.albums); //set the state for wishlist albums
+        // Store each album id from wishlist in local storage
+        wishlistData.albums.forEach((album) => {
+          localStorage.setItem(
+            `wishListAlbums-${album._id}`,
+            JSON.stringify(true)
+          );
+        });
```
akosasante: Same suggestion as in `AlbumsList` about potentially updating this to store one single entry in localstorage with an object that has all the wishlist albums, rather than a separate entry for each wishlist album.

src/components/wishList/WishListView.js
```diff
@@ -0,0 +1,35 @@
...
+      } catch (error) {
+        console.error('Error fetching wishlist:', error);
+      }
+    };
+
+    fetchWishlist();
+  }, []);
+
+  return <AlbumGrid albums={albums} wishListId={wishListId} />;
```
akosasante: Currently our wishlist page looks relatively blank; may be worthwhile to add a container to wrap the `AlbumGrid` and maybe a header to indicate that this is the wishlist page.

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -4,49 +4,39 @@ import axiosInstance from '../../../apis/axiosClient'; // axios instance
 import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
 import { useNavigate } from 'react-router-dom';
 
-const wishlistEndpoint = '/wishlist';
```
akosasante, line 7: This deleted line is making the api calls for addToWishlist and removeFromWishlist erorr out currently. Either restore this line or update those api calls so that they just hardcode the string `/wishlist`

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -4,49 +4,39 @@ import axiosInstance from '../../../apis/axiosClient'; // axios instance
...
-      return response.data.wishlist._id; // Return the wishlist ID
-    } catch (error) {
-      console.error('Error creating wishlist:', error);
-    }
-  };
-
   useEffect(() => {
     const storedAddedStatus =
-      JSON.parse(localStorage.getItem(`addedStatus-${album._id}`)) || false;
+      JSON.parse(localStorage.getItem(`wishListAlbums-${album._id}`)) || false;
```
akosasante: If you do end up making the localstorage field a single object, then this will need to be updated to take that into account:

```suggestion
    const storedWishlist = JSON.parse(localStorage.getItem('wishlistAlbums')) || {};
    const storedAddedStatus = album._id in storedWishlist;
```

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -4,49 +4,39 @@ import axiosInstance from '../../../apis/axiosClient'; // axios instance
...
 
   const addToWishlist = async () => {
-    const wishlistId = await fetchOrCreateWishlist(); // Fetch the wishlist ID
     try {
       await axiosInstance.patch(
-        `${wishlistEndpoint}/${wishlistId}/add_album/${album._id}`
+        `${wishlistEndpoint}/${wishListId}/add_album/${album._id}`
       );
-      localStorage.setItem(`addedStatus-${album._id}`, JSON.stringify(true));
+      localStorage.setItem(`wishListAlbums-${album._id}`, JSON.stringify(true));
```
akosasante: If you end up making the localstorage field an object, this will need to be updated to take that into account:

```suggestion
      const wishlistAlbums = JSON.parse(localStorage.getItem('wishlistAlbums'))
      wishlistAlbums[album._id] = album;
      localStorage.setItem("wishlistAlbums", JSON.stringify(wishlistAlbums));
```

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -4,49 +4,39 @@ import axiosInstance from '../../../apis/axiosClient'; // axios instance
...
-    const wishlistId = await fetchOrCreateWishlist(); // Fetch the wishlist ID
     try {
       await axiosInstance.patch(
-        `${wishlistEndpoint}/${wishlistId}/remove_album/${album._id}`
+        `${wishlistEndpoint}/${wishListId}/remove_album/${album._id}`
+      );
+      localStorage.setItem(
+        `wishListAlbums-${album._id}`,
+        JSON.stringify(false)
       );
```
akosasante: If you end up making the localstorage field an object, this will need to be updated to take that into account:

```suggestion
      const wishlistAlbums = JSON.parse(localStorage.getItem('wishlistAlbums'))
      delete wishlistAlbums[album._id];
      localStorage.setItem("wishlistAlbums", JSON.stringify(wishlistAlbums));
```

akosasante: You may want to have the page either manually reload (with react-router this can be done with `const navigate = useNavigate();` and then where you wan to refresh `navigate(0);`. 

This will ensure that when you remove an item from the wishlist, it goes away from the wishlist view.

Alternatively, you could pass down an `onRemove` function from the `WishlistView` (it would have to be drilled through the AlbumGrid and Album components) that will tell the WishlistView, which album id was removed, and then in WishlistView we can call `setAlbums` to update the list of albums to be shown on the grid (the actual API call to remove from wishlist would still remain here in the button component). 

A third possible option for doing this is using a Redux action so that you don't have to drill the prop through multiple components.

src/components/wishList/WishListView.js
akosasante: You may also want to have some sort of empty state message when the user's wishlist has no albums on it.

akosasante (commented):
We may want to add a button in the navbar to allow the user to easily navigate to their wishlist when they are logged in.

src/components/Album.js
JafariM: Done

src/components/AlbumsList.js
JafariM: I changed it to object

src/components/wishList/WishListView.js
JafariM: Done

src/components/AlbumsList.js
JafariM: Removed

src/components/wishList/WishListView.js
JafariM: Done

src/components/wishList/WishListView.js
JafariM: Done

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
JafariM: Sorry, I accidently removed that line after I checked that addToWishlist and removeFromWishlist works good, I restored taht

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
JafariM: Done

src/components/Navbar.js
```diff
@@ -34,6 +34,9 @@ const Navbar = () => {
               Profile
             </Button>
             <Logout />
+            <Button color="inherit" component={Link} to="/wishlist">
+              Wishlist
+            </Button>
```
akosasante: Not a huge deal, but usually websites have the `Logout` button as the last (right-most) entry in the navbar so it's easy to find.

src/components/wishList/WishListView.js
```diff
@@ -0,0 +1,62 @@
+import React, { useState, useEffect } from 'react';
+import AlbumGrid from '../AlbumGrid';
+import axiosInstance from '../../apis/axiosClient';
+
+import Container from '@mui/material/Container';
+import Grid from '@mui/material/Grid';
```
akosasante: According to the lint message from Github Actions we can remove this line now that it's no longer used

akosasante (approved):
Looking and working awesomely!

src/components/Navbar.js
JafariM: You are right, I moved the logout to end

PR #46, opened 2023-09-09
PS-197 Album marketplace pagination

AlonaVladymyrovaTrinity:
**Description:**
Added album marketplace pagination for improved user experience.

**Testing Steps:**

- Navigate to the album marketplace page [http://localhost:3000/home](http://localhost:3000/home) in the web application.

- Verify that the pagination UI is displayed at the bottom of the album list.

**Test the following scenarios:**

1. Click on different page numbers in the pagination component to switch between pages.
2. Verify that the album list updates accordingly based on the selected page.
3. Perform searches using different search terms and search types and verify that pagination remains functional.
4. Ensure that the pagination feature does not introduce any console errors or warnings.

Screenshot 2023-09-09 at 1 10 39 AM


akosasante (commented):
Looking good! A few issues I'm running into are:

**(1)**

*   navigate to some page (eg: page 3)
*   type in a search term that I know should have a result (eg: "Album Name" = "Love")
*   click "Search" button and receive no results
*   check the network tab, it looks like the payload that was sent to the backend was: `limit=12&albumName=Love&offset=24`. That means it's asking the backend for the 3rd page of results that match the album name "Love". But since that search term only returns 1 album then there's nothing to show on the 3rd page.

So what's needed is to ensure that when a new search is submitted, we reset the currnet page number to 1 before making the request; as part of `handleSearch`. 

**(2)**

The other thing is that after putting in a search term, the page numbers at the bottom should reflect how many results match that search term. This may require a change on the backend to return the `total` number of results (since the `count` field that's returned right now only indicates the number of results on that _page_). 

**(3)**

The last thing I noticed is that if I have searched for an album term (eg: "Album Name": "Love") and then I try to page through the results, the subsequent requests will update the `offset` query parameter sent to the backend, but the `albumName` query parameter is set to a blank string due to this line where in the `useEffect` that gets triggered by the page change, we pass in an empty string as the 'searchTerm' given to the fetchAlbums function: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/PS-197-Album-marketplace-pagination/src/components/AlbumsList.js#L64-L66](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/PS-197-Album-marketplace-pagination/src/components/AlbumsList.js#L64-L66)

I think one approach for resolving **(1)** and **(3)** would be to have a separate state variable that's used to store the last submitted search term and using that value as what's passed  into/used by the fetchAlbums function and the useEffect that calls it.

AlonaVladymyrovaTrinity:
I tried to fix the first and third points. Please review. 
I believe that Back End is to busy right now with presentation preparation. I don't think that they could add something at Back End in nearest future.

src/components/album/AlbumsList.js
```diff
@@ -84,7 +92,9 @@ const AlbumsList = () => {
 
   //call the function to make API request for search input value
   const handleSearch = () => {
-    fetchAlbums(searchType, searchTerm, limit);
+    setLastSearchTerm(searchTerm);
+    setCurrentPage(1);
+    fetchAlbums(searchType, searchTerm);
   };
 
   //clear search input and make an empty API call
```
akosasante, line 100: I think the below is from before your changes, but I just noticed when testing things out on this branch. We should not be setting `limit` to `1` when pressing the `Clear` button (line 104). We should also ensure that `fetchAlbums` is being called with a valid page numb er (line 106; I think we want to call `fetchAlbums('albumName', '', 1)` instead of passing it `0` which causes offset to become a negative number).

akosasante: Actually, we can remove line 106 altogether. We don't need to explicitly call `fetchAlbums`, it will already get called as a result of `setSearchTerm('')` changing the searchTerm, which will trigger the `useEffect` hook on line 65 which runs the fetchAlbum call any the search term or the search type change

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,47 @@ import {
...
+    },
+    [/*currentPage,*/ limit]
+  );
 
-  //display 10 albums when first user visit this page
   useEffect(() => {
-    fetchAlbums('albumName', '', 10);
-  }, []);
+    fetchAlbums(searchType, searchTerm); // Use current search term
+  }, [fetchAlbums, searchType, searchTerm]); // Refetch albums when the page changes
```
akosasante: This will run the `fetchAlbum` call, any time the user types into the search box, even before they click the "Search" button. This is fine, but if that's the case then it seems like having the search button is a bit redundant. Either we remove the search button, or we update this useEffect to use `lastSearchTerm` instead:

```suggestion
    fetchAlbums(searchType, lastSearchTerm); // Use current search term
  }, [fetchAlbums, searchType, lastSearchTerm]); // Refetch albums when the submitted search changes
```

If we do that, we should make sure to update `handleClear` to also setLastSearchTerm to an empty string, so that the fetchAlbum call is triggered in that case as well

src/components/album/AlbumsList.js
```diff
@@ -84,7 +92,9 @@ const AlbumsList = () => {
 
   //call the function to make API request for search input value
   const handleSearch = () => {
-    fetchAlbums(searchType, searchTerm, limit);
+    setLastSearchTerm(searchTerm);
+    setCurrentPage(1);
+    fetchAlbums(searchType, searchTerm);
```
akosasante: If we update the `useEffect` on line 66 to be dependent on the `lastSearchTerm` as suggested in above comment, then we can remove this line( 97) as won't need to explicitly call `fetchAlbums` here, it will just happen by triggering that useEffect call.

akosasante: If we update the `useEffect` on line 66 to be dependent on the `lastSearchTerm` as suggested in above comment, then we won't need to explicitly call `fetchAlbums` here, it will just happen by triggering that useEffect call.

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,47 @@ import {
 
 const AlbumsList = () => {
   const [albums, setAlbums] = useState([]); // for albums
-  const [limit, setLimit] = useState(1); // for pagination
+  const [limit, setLimit] = useState(12); // for pagination
```
akosasante: Since we never need to change the `limit` / page size, we can probably make this a static variable instead of a part of the component state (unless we forsee some use for it in the future in which case Im fine to leave it as is)

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: Code fixed according to them recommendations

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: Code fixed according to them recommendations

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: Code fixed according to them recommendations

AlonaVladymyrovaTrinity: Code fixed according to them recommendations

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: Code fixed according to them recommendations

AlonaVladymyrovaTrinity:
Akos, I made code fixes according to your recommendations, but I believe that 'CLEAR' search button is not working now as expected. Would you take a look and give me some more advice on how to fix this? Thank you!

AlonaVladymyrovaTrinity:
> I think the below is from before your changes, but I just noticed when testing things out on this branch. We should not be setting limit to 1 when pressing the Clear button (line 104). We should also ensure that fetchAlbums is being called with a valid page numb er (line 106; I think we want to call fetchAlbums('albumName', '', 1) instead of passing it 0 which causes offset to become a negative number).
> Actually, we can remove line 106 altogether. We don't need to explicitly call fetchAlbums, it will already get called as a result of setSearchTerm('') changing the searchTerm, which will trigger the useEffect hook on line 65 which runs the fetchAlbum call any the search term or the search type change

Maybe I shouldn't remove this line `fetchAlbums('albumName', '', 1);`  in the function:
```
 //clear search input and make an empty API call
  const handleClear = () => {
    setSearchType('albumName');
    setSearchTerm('');
    // setLimit(1);
    setMessage('');
    // fetchAlbums('albumName', '', 1);  
  };
```
What do you think?

src/components/album/AlbumsList.js
```diff
@@ -170,13 +186,32 @@ const AlbumsList = () => {
...
+                sx={{
+                  display: 'flex',
+                  justifyContent: 'center',
+                  mt: '3rem',
+                  mb: '2rem',
+                }}
+              >
+                <Stack spacing={2}>
+                  <Pagination
+                    count={6} // Replace with the total number of pages (on this moment we have 69 albums / 12 albums per page)
```
akosasante: Recently, the backend has deployed a change that returns additional fields in the `filter albums` endpoint: Code-the-Dream-School/dd-prac-team4-back#74



So that means now we can have a state variable for `totalPages` and pass that in here instead of a static value of `6`. That way the correct number of pages will be shown based on how many results are returned for the given query :smiley:

src/components/album/AlbumsList.js
```diff
@@ -84,16 +95,17 @@ const AlbumsList = () => {
...
   //clear search input and make an empty API call
   const handleClear = () => {
     setSearchType('albumName');
     setSearchTerm('');
-    setLimit(1);
+    // setLimit(1);
     setMessage('');
-    fetchAlbums('albumName', '', 0);
+    // fetchAlbums('albumName', '', 1);
   };
```
akosasante: The `useEffect` that I'd like to rely on to call `fetchAlbums`, currently has the following dependencies: `[fetchAlbums, searchType, lastSearchTerm]`. That means it will be triggered if any of those three variables is updated. 

Currently, in `handleClear` we are not updating any of those. We have `setSearchTerm('')` which is responsible for clearing the input in the UI. But we should _**also**_ have `setLastSearchTerm('');` which will trigger the useEffect and will represent the value we send in the API call to an empty string. 

---

In addition, some things to consider, do we need to specifically reset the `searchType` here when the user clicks "Clear"? It may be a bit surprising if the user has been searching "Artists", that the Clear button would switch that back to searching by "Album". 

Similarly, is there a case where we specifically need to update the `message` variable only when Clearing, but we don't need to do it when searching? It seems to me like we may be able to remove the `setMessage` line as well.

src/components/album/AlbumsList.js
```diff
@@ -120,6 +132,10 @@ const AlbumsList = () => {
     </React.Fragment>
   );
   //end of snackbar
+  const handlePageChange = (event, page) => {
+    setCurrentPage(page);
+    fetchAlbums(searchType, lastSearchTerm, page);
+  };
 
   return (
     <Container>
```
akosasante, line 146: On line 157, we are setting the search type when someone changes the value in the dropdown. 

One side effect of this is that it will trigger the `useEffect` on  line 57 to call `fetchAlbums`, with the same current page that we currently are on. That may cause some oddness if the user has already paginated through their current search term before deciding to switch search types:


https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/b0fc87e8-6aec-45b1-93db-60afb3215a62



We can avoid this by instead creating a handler function `handleSearchTypeChange` that takes this click event and sets the `currentPage` back to `1` _in addition to_ setting the `searchType` variable.

akosasante (commented):
Hi Alona :wave: :smile:

akosasante:
@AlonaVladymyrovaTrinity  are you looking to continue working on / finishing up this PR or should we merge it as-is for now and anyone continuing on this repo can do some follow up work to resolve the few remaining loose ends?

AlonaVladymyrovaTrinity:
@akosasante Hi Akos, thanks for reaching out regarding this issue! Just wanted to let you know I've pushed two commits today.  If anyone in the team has ideas on how to solve this issue by using a different approach, feel free to assign this ticket to another person. I also don't mind if someone wants to change, improve or optimize my code — I'd just be glad if my work proves helpful. Thank you! And happy upcoming holidays!

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: Did that in [31ab91f](https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/46/commits/31ab91f9f472c0ab2f942c5c353db6dc338f6b95)

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: [31ab91f](https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/46/commits/31ab91f9f472c0ab2f942c5c353db6dc338f6b95)

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,51 @@ import {
...
-        `${process.env.REACT_APP_API_BASE_PATH}/albums/filter`,
-        {
-          params: {
-            limit,
-            [searchType]: searchTerm,
-          },
+  const fetchAlbums = useCallback(
+    async (searchType, searchTerm) => {
+      try {
+        const offset = (searchTerm === '' ? currentPage - 1 : 0) * limit;
```
akosasante: Ohh nice, I like the idea of this approach!

We do again have an issue where, if I have a search term, then pagination doesn't work:

Eg:


https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/07def0b6-b71e-4dd3-89ef-da82a4112e21


*   Click next, `currentPage` is updated to `2`. `searchTerm` is still `"L"`. `offset = ("L" === "" ? 2 -1 : 0)*12 = (false ? 1 :0)*12 = (0) * 12 = 0`

src/components/album/AlbumsList.js
```diff
@@ -170,13 +185,32 @@ const AlbumsList = () => {
...
+                sx={{
+                  display: 'flex',
+                  justifyContent: 'center',
+                  mt: '3rem',
+                  mb: '2rem',
+                }}
+              >
+                <Stack spacing={2}>
+                  <Pagination
+                    count={6} //{albums.totalPages}
```
akosasante: In the `fetchAlbums` we get the `albums` part of the response, and are setting a state variable. We can do the same with `totalPages`

something like:

```
const [totalPages, setTotalPages] = useState(1); // Track total pages for this search

// in fetchAlbums, set the value of total pages using response.data.totalPages

// here in the Pagination component, set the `count` attribute to totalPages
```

src/components/album/AlbumsList.js
akosasante: I think we can instead always, update the offset based on the current page, regardless of the searchTerm value:

```suggestion
        const offset = (currentPage - 1) * limit;
```

src/components/album/AlbumsList.js
```diff
@@ -84,16 +96,15 @@ const AlbumsList = () => {
 
   //call the function to make API request for search input value
   const handleSearch = () => {
-    fetchAlbums(searchType, searchTerm, limit);
+    fetchAlbums(searchType, searchTerm);
   };
```
akosasante: Rather than calling `fetchAlbums` ourselves, we can just update the `currentPage` to `1`. This will trigger the useEffect on line 69, which will call fetchAlbums, with the appropriate searchType and searchTerm values (based on my review comment left on that line).

```suggestion
  const handleSearch = () => {
    if (currentPage === 1) {
      // resetting the current page to 1 won't trigger useEffect so here we have to call the fetch callback manually
      fetchAlbums(searchType, searchTerm);
    } else {
      setCurrentPage(1);
    }
  };
```

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,51 @@ import {
...
-  //display 10 albums when first user visit this page
+    },
+    [currentPage, limit]
+  );
+  //display 12 albums when first user visit this page
   useEffect(() => {
-    fetchAlbums('albumName', '', 10);
-  }, []);
+    fetchAlbums('albumName', '');
+  }, [fetchAlbums, currentPage]); // Refetch albums when the page changes
```
akosasante:
```suggestion
  useEffect(() => {
    fetchAlbums(searchType, searchTerm);
  }, [fetchAlbums, currentPage]); // Refetch albums when the page changes
```

This will ensure that we refetch, with the latest values of searchType and searchTerm

src/components/album/AlbumsList.js
```diff
@@ -84,16 +96,15 @@ const AlbumsList = () => {
...
   //clear search input and make an empty API call
   const handleClear = () => {
-    setSearchType('albumName');
+    //setSearchType('albumName');
     setSearchTerm('');
-    setLimit(1);
     setMessage('');
-    fetchAlbums('albumName', '', 0);
+    fetchAlbums('albumName', '');
   };
```
akosasante: This approach is fine, but we also want to make sure that we go back to page 1 when we've cleared the search so that we're getting the fresh/blank search term results. Changing current page will call `fetchAlbums` for us via the useEffect on line 69:

```javascript
  const handleClear = () => {
    setSearchTerm('');
    setMessage('');
    setCurrentPage(1);
  };
```

src/components/album/AlbumsList.js
```diff
@@ -121,6 +132,10 @@ const AlbumsList = () => {
   );
   //end of snackbar
 
+  const handlePageChange = (event, page) => {
+    setCurrentPage(page);
+  };
```
akosasante, line 136: :white\_check\_mark:

akosasante (commented):
Great work Alona; I think this last set of fixes should be good to have us ready to merge!

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: I was not able to play the video.
Anyway I did this fix and unfortunately after that the search is stoped working at all from all pages except first. So, if you are on the first page and you are searching for any albums or artists(from any page) - it works, but the search from other pages - not working at all in this case. 
So, should I apply this fix, or it will be better to leave it as is?

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: I applied this change. Thank you for detailed explanation. Now I understand! :)

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: After applying this fix, if the user is trying to do search from any page (except the first one) - after first click on SEARCH button he is redirected to the first page, and then he need to press the SEARCH button again to get a search result. We don't have a written requirements, so if this is an expected behavior, I can apply it.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: I did this, but now I got this error:
  Line 80:6:  React Hook useEffect has missing dependencies: 'searchTerm' and 'searchType'.

AlonaVladymyrovaTrinity: I added them to a dependency list:   
`}, [fetchAlbums, currentPage, searchTerm, searchType]);
`

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: I added these changes

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: After adding all the changes listed below, this fix began to work.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: After adding all the changes, this fix began to work. However, in this case we are not using the Search button at all! Should we keep it?

AlonaVladymyrovaTrinity:
I have made all the changes according to your latest recommendations. Please review my recent changes again.

src/components/album/AlbumsList.js
akosasante: Looks good; that is the correct next step there 👍 .

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,60 @@ import {
...
+  const [totalPages, setTotalPages] = useState(1); // Track total pages for this search
+  const limit = 12;
+  // const [limit, setLimit] = useState(12); // for pagination
   const [searchType, setSearchType] = useState('albumName'); //default value for select input
   const [searchTerm, setSearchTerm] = useState(''); // for search input filed value
   const [message, setMessage] = useState(''); // for not found message
   const [errorMessage, setErrorMessage] = useState(''); // for error message
   const [wishListId, setWishListId] = useState();
+  const [currentPage, setCurrentPage] = useState(1); // Track current page
+  // const [lastSearchTerm, setLastSearchTerm] = useState('');
```
akosasante:
```suggestion
  const [albums, setAlbums] = useState([]); // for albums
  const [totalPages, setTotalPages] = useState(1); // Track total pages for this search
  const [searchType, setSearchType] = useState('albumName'); //default value for select input
  const [searchTerm, setSearchTerm] = useState(''); // for search input filed value
  const [message, setMessage] = useState(''); // for not found message
  const [errorMessage, setErrorMessage] = useState(''); // for error message
  const [wishListId, setWishListId] = useState();
  const [currentPage, setCurrentPage] = useState(1); // Track current page
  const [lastSearchTerm, setLastSearchTerm] = useState(''); // Track the input value of the search last time the user clicked "Search" or "Clear" button
  // Page limit is static/won't change across re-renders, so we don't need to set it in component state
  const limit = 12;
```

akosasante: This suggestion adds a `lastSearchTerm` state variable (I believe you had something like this before -- we'll want to keep it if we go the route of having to click the "Search" button explicitly. In that case we want to have a variable where we can store the last value of the search term rather than the live value as the user is typing into the input box); it also moves `limit` to the bottom just for organizational purposes.

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,60 @@ import {
...
-      setOpen(true);
-    }
-  };
+    },
+    [currentPage, limit]
+  );
+  //display 12 albums when first user visit this page
+  // useEffect(() => {
+  //   fetchAlbums('albumName', '');
+  // }, [fetchAlbums, currentPage]); // Refetch albums when the page changes
```
akosasante:
```suggestion
  //make an API call with search values to backend and return the result
  const fetchAlbums = useCallback(
    // We don't need to pass any values to this function because it will make use of the component state
    async () => {
      try {
        // Calculate the offset, based on the current page (alternatively, we could use the `nextPage`/`prevPage` values returned by the API; but this is more flexible since the user can also clear the search or switch search type altogether)
        const offset = (currentPage - 1) * limit;

        //for testing
        // console.log(currentPage - 1 + ' * ' + limit + ' = ' + offset);
        // console.log('searchType:' + searchType + ' ; searchTerm:' + lastSearchTerm);

        const response = await axios.get(
          `${process.env.REACT_APP_API_BASE_PATH}/albums/filter`,
          {
            params: {
              limit: limit,
              [searchType]: lastSearchTerm,  // we use the `lastSearchTerm` value which is only set by us explicitly in code, rather than `searchTerm` which would also get changed whenever the user types into the input box
              offset,
            },
          }
        );

        const searchResult = response.data.albums;
        setAlbums(searchResult);
        setTotalPages(response.data.totalPages);

        if (searchResult.length === 0) {
          setMessage('No result found');
        }
      } catch (error) {
        console.error('Error fetching albums:', error);
        setErrorMessage(error.message);
        setOpen(true);
      }
    },
    [currentPage, lastSearchTerm, searchType]
  );
```

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,60 @@ import {
...
 
-  //display 10 albums when first user visit this page
   useEffect(() => {
-    fetchAlbums('albumName', '', 10);
-  }, []);
+    fetchAlbums(searchType, searchTerm);
+  }, [fetchAlbums, currentPage, searchTerm, searchType]);
```
akosasante:
```suggestion
  // This will run on component mount and whenever the `fetchAlbums` function changes
  // Based on the dependency array we gave the `useCallback` above which wraps `fetchAlbums`,
  // this means that this `useEffect` will run any time the `currentPage` or `lastSearchTerm` or `searchType` values change
  useEffect(() => {
    fetchAlbums();
  }, [fetchAlbums]);
```

src/components/album/AlbumsList.js
```diff
@@ -84,16 +105,23 @@ const AlbumsList = () => {
...
   const handleSearch = () => {
-    fetchAlbums(searchType, searchTerm, limit);
+    //   fetchAlbums(searchType, searchTerm);
+
+    if (currentPage === 1) {
+      // resetting the current page to 1 won't trigger useEffect so here we have to call the fetch callback manually
+      fetchAlbums(searchType, searchTerm);
+    } else {
+      setCurrentPage(1);
+    }
```
akosasante:
```suggestion
  //call the function to make API request for search input value
  const handleSearch = () => {
    if (currentPage !== 1) {
      // If the user is on a page other than the first page, we want to reset to the first page so that the new search results are displayed from the beginning
      setCurrentPage(1);
    }
    // Set "lastSearchTerm" to the current value in the input; this will trigger the useEffect above to call fetchAlbums
    setLastSearchTerm(searchTerm);
  };
```

src/components/album/AlbumsList.js
```diff
@@ -84,16 +105,23 @@ const AlbumsList = () => {
...
   const handleClear = () => {
-    setSearchType('albumName');
+    //setSearchType('albumName');
     setSearchTerm('');
-    setLimit(1);
     setMessage('');
-    fetchAlbums('albumName', '', 0);
+    setCurrentPage(1);
+    // fetchAlbums('albumName', '');
   };
```
akosasante:
```suggestion
  //clear search input and make an empty API call
  const handleClear = () => {
    setSearchTerm('');
    setMessage('');
    setCurrentPage(1);
    setLastSearchTerm('');
    // Because we've updated the `lastSearchTerm`, and `currentPage` values, this will trigger the useEffect above to call fetchAlbums
  };
```

src/components/album/AlbumsList.js
```diff
@@ -121,6 +149,10 @@ const AlbumsList = () => {
   );
   //end of snackbar
 
+  const handlePageChange = (event, page) => {
+    setCurrentPage(page);
+  };
```
akosasante, line 136:
```suggestion
  const handlePageChange = (event, page) => {
    // Changing the current page will update the UI and will also trigger the useEffect above to call the `fetchAlbums` callback
    setCurrentPage(page);
  };

  const handleSearchTypeChange = (event) => {
    setSearchType(event.target.value);
    setSearchTerm('');
    setCurrentPage(1);
    setLastSearchTerm('')
    // Because we've updated the `searchType`, `currentPage` and `lastSearchTerm` values, this will trigger the useEffect above to call fetchAlbums
  };
```

akosasante: This would need to be accompanied with updating line 185 to `onChange={handleSearchTypeChange}`

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,60 @@ import {
...
+
+        //for testing
+        // console.log(currentPage - 1 + ' * ' + limit + ' = ' + offset);
+        // console.log('searchType:' + searchType + ' ; searchTerm:' + searchTerm);
+
+        const response = await axios.get(
+          `${process.env.REACT_APP_API_BASE_PATH}/albums/filter`,
+          {
+            params: {
+              limit: limit,
```
akosasante: Can be shortened to

```suggestion
              limit,
```

src/components/album/AlbumsList.js
```diff
@@ -22,41 +24,60 @@ import {
...
+      try {
+        //const offset = (currentPage - 1) * limit;
+        const offset = (searchTerm === '' ? currentPage - 1 : 0) * limit;
+
+        //for testing
+        // console.log(currentPage - 1 + ' * ' + limit + ' = ' + offset);
+        // console.log('searchType:' + searchType + ' ; searchTerm:' + searchTerm);
+
+        const response = await axios.get(
+          `${process.env.REACT_APP_API_BASE_PATH}/albums/filter`,
```
akosasante: We should prefer in this repo to use `axiosInstance` from the `src/apis/axiosClient` folder; because it sets cookies and the base url for us already:

```suggestion
        const response = await axiosInstance.get('/albums/filter',
```

akosasante (commented):
Sorry for the long delay between reviews; I've been neck-deep in work :cry: 

I think we've been circling around a solution here that will fix all issues. I'm happy with sticking with the approach you have come to which allows searching without clicking the "Search" button explicitly. If we go with that; we should probably remove the `"Search"` button to reduce any confusion then. 

I've also left comments with suggestions for other approach where focus on letting `useEffect` and `useCallback` dependencies drive the pagination.

I'm happy to move forward with either approach, please feel free to pick the one that appeals to you or feels most maintainable/understandable.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: > Sorry for the long delay between reviews; I've been neck-deep in work 😢
No worries. There is no rush anyway…
These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

src/components/album/AlbumsList.js
AlonaVladymyrovaTrinity: These changes have been applied.

AlonaVladymyrovaTrinity:
Hi, Akos @akosasante ! I applied all the changes you suggested above and now everything works as expected, please review this PR again. Thank you!

akosasante (approved):
Fantastic job! Tested the various cases that we've brought up throughout this PR and everything is working swimmingly. Thanks so much for your patience and your perseverance. Really appreciate it!

Since it's been a long-lived branch there _may_ be some conflicts to resolve when merging, but hopefully it'll be pretty straightforward. Let me know in Slack if you run into any issues.

AlonaVladymyrovaTrinity:
Thank You so much for all Your support and mentorship during all practicum and even after it. I really greatly appreciate this. It was a great pleasure for me to have Your knowledge, advices and experience behind “my back”. For myself, it was a perfect chance to study from so such a great mentor.
Have a happiness holiday time ever!

PR #47, opened 2023-09-09
Create icons for wishlist and cart buttons

JafariM:


akosasante (approved):
Looks great, thanks for picking this up!

One enhancement we could make is for even users who don't need screen readers, it may not be obvious to _all_ people what the icons mean. We could wrap the icon button elements in a Tooltip element ([basic tooltip docs in material-ui](https://mui.com/material-ui/react-tooltip/#basic-tooltip)) so that when a user hovers over the icon, it says what it does (eg: "remove from wishlist").

Approving cause this works well here but you can decide if you want to add the tooltip here or leave it for some later point.

JafariM:
> Looks great, thanks for picking this up!
> 
> One enhancement we could make is for even users who don't need screen readers, it may not be obvious to _all_ people what the icons mean. We could wrap the icon button elements in a Tooltip element ([basic tooltip docs in material-ui](https://mui.com/material-ui/react-tooltip/#basic-tooltip)) so that when a user hovers over the icon, it says what it does (eg: "remove from wishlist").
> 
> Approving cause this works well here but you can decide if you want to add the tooltip here or leave it for some later point.

I add totoltip to buttons

PR #49, opened 2023-09-19
forum.js roomJoin.js

mvstoyan:
Forum page
Create rooms

### DIFF
--- a/src/components/forum/RoomJoin.js
+++ b/src/components/forum/RoomJoin.js
@@ -1,0 +1,13 @@
+import React from 'react';
+
+const RoomJoin = ({ handleJoinRoom }) => {
+  return (
+    <div>
+      <input type="text" placeholder="Room ID" />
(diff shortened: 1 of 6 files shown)


src/components/forum/RoomJoin.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const RoomJoin = ({ handleJoinRoom }) => {
+  return (
+    <div>
+      <input type="text" placeholder="Room ID" />
+      <input type="text" placeholder="User ID" />
```
akosasante, line 7: Would the user have to know their own user id and know the correct room id to input them here? I think maybe that's something we can handle for them since we know their user id from them being logged in; and I assume we know the room id based on which "Join Room" button they are clicking?

akosasante:
Per our chat in Slack, closing this PR for now (can always be reopened in the future)

PR #50, opened 2023-09-24
Album chat page

mvstoyan:


src/components/album/AlbumChat.js
```diff
@@ -0,0 +1,35 @@
+import React, { useEffect } from 'react';
+import io from 'socket.io-client';
+
+const AlbumChat = ({ apiUrl }) => {
+  useEffect(() => {
+    const socket = io('http://localhost:8000');
```
akosasante: Just like the HTTP API client in the frontend, we should use an environment variable so that this points to localhost when running locally and [https://musicstore.onrender.com](https://musicstore.onrender.com) when running in production (eg: the API uses `REACT_APP_API_BASE_PATH`, here we could use `REACT_APP_SOCKET_BASE_PATH`)

src/components/album/AlbumPreview.js
```diff
@@ -31,6 +32,8 @@ const AlbumPreview = ({
   }, [apiUrl]);
 
   return (
+    <div>
+      <AlbumChat />
```
akosasante: Your `AlbumChat.js` component is currently expecting to be passed an `apiUrl` but here we're not passing it any props, so currently when that component loads the value will be undefined. If we want to pass the prop in we need to define it like an HTML attribute:

```suggestion
      <AlbumChat apiUrl={apiUrl} />
```

akosasante (commented):
quick drive-by review: 

You may need to add the CORS settings from [https://codethedreamteam4-2023.atlassian.net/browse/PS-186](https://codethedreamteam4-2023.atlassian.net/browse/PS-186) to the backend in order to be able to test these frontend changes.

src/components/album/AlbumPreview.js
akosasante: And if you're looking for a way to test this locally, you can try creating a temporary route that loads this `AlbumChat` component with a static `apiUrl`

eg in App.js:

```js
import AlbumChat from './components/album/AlbumChat';

<Routes>
 ...all the existing routes
  <Route path="/chat" element={<AlbumChat apiUrl="https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy" />} />
</Routes>
```

(I just picked a random album from the db)

then you can navigate to `localhost:3000/chat` to view your component.

--- 

One thing to keep in mind, is that the album id in the Spotify URL is _Spotify's_ album ID; it won't match the Mongo Object Id that we identify the album by in our db. This may not matter if we're not using this id for anything other than naming the chat channel but just wanted to point it out in case you weren't aware.

src/components/album/AlbumChat.js
mvstoyan: .env: REACT_APP_SOCKET_BASE_PATH=http://localhost:8000
REACT_APP_SOCKET_BASE_PATH= https://musicstore.onrender.com

screen

src/components/album/AlbumPreview.js
mvstoyan: Thank you!!

.env
```diff
@@ -1 +1,3 @@
-REACT_APP_API_BASE_PATH=http://localhost:8000/api/v1
\ No newline at end of file
+REACT_APP_API_BASE_PATH=http://localhost:8000/api/v1
+REACT_APP_SOCKET_BASE_PATH=http://localhost:8000
+REACT_APP_SOCKET_BASE_PATH= https://musicstore.onrender.com
```
akosasante: We only need to have our dev variables here in the \`.env\` file.

By including \`REACT\_APP\_SOCKET\_BASE\_PATH\` twice here, the second value will overwrite the first one. So when you run your app locally, it will be trying to connect to the production server:



For production, just like with our backend server, it doesn't use the `.env` file. Rather, it will be getting the environment variables from those injected by Render.com when the app is started. We provide those variables in the Render.com settings pane for the frontend app: https://dashboard.render.com/static/srv-cjf0j06e546c73d6lq10/env

src/App.js
```diff
@@ -31,6 +32,12 @@ function App() {
         <Route path="/checkout/completed" element={<CheckoutComplete />} />
         <Route path="/profile" element={<UserProfile />} />
         <Route path="*" element={<PageNotFound />} />
+        <Route
+          path="/chat"
+          element={
+            <AlbumChat apiUrl="https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy" />
```
akosasante: In fact, maybe it's even better to use `AlbumPreview` as the component you're testing, just to make sure that the chat component you've nested in there looks good / works well altogether 

```suggestion
            <AlbumPreview apiUrl="https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy" />
```

src/App.js
```diff
@@ -10,6 +10,7 @@ import WishListView from './components/wishList/WishListView';
 import UserProfile from './components/user/UserProfile';
 import CheckoutComplete from './components/Purchase/CheckoutComplete';
 import PageNotFound from './components/PageNotFound';
+import AlbumChat from './components/album/AlbumChat';
```
akosasante:
```suggestion
import AlbumPreview from './components/album/AlbumPreview';
```

src/components/album/AlbumChat.js
```diff
@@ -0,0 +1,37 @@
...
+    const match = apiUrl.match(regexPattern);
+    if (match) {
+      // Extract albumId from the URL
+      const albumId = match[1];
+      // Сonnection is established with the Socket.io server
+      socket.on('connect', () => {
+        // Request to the server to join the album chat
+        console.log('Connected to WebSocket server');
+        socket.emit('join:album_chat', albumId);
+      });
```
akosasante: Looking good! Last step to do based on ticket PS-184 is to add a step where you subscribe to the `chat:album` channel (In the ticket I previously had `chat:album:<albumId>` but that was incorrect, ticket has been fixed now). You'll want that nested here inside the \`on connect\`. For now the listener function for that could just log something to the console; we'll expand on it in the subsequent tickets

.env
mvstoyan: Removed. I see now, we have on render this, thank you for explanation.

akosasante (approved):

PR #51, opened 2023-09-29
PS185-send-chat-messages

mvstoyan:


mvstoyan:
screen
have 404 bad request and cant fix(( .env REACT_APP_SOCKET_BASE_PATH=http://localhost:8000/api/v1

src/components/album/AlbumChat.js
```diff
@@ -1,23 +1,32 @@
...
+import React, { useEffect, useState } from 'react';
 import io from 'socket.io-client';
 
 const AlbumChat = ({ apiUrl }) => {
-  useEffect(() => {
+  const [message, setMessage] = useState('');
+
+  // Nandle sending a chat message
+  const handleSendMessage = () => {
     const socket = io(process.env.REACT_APP_SOCKET_BASE_PATH);
```
akosasante: Oh one thing I missed earlier; is that `REACT_APP_SOCKET_BASE_PATH` should be `http://localhost:8000` not `http://localhost:8000/api/v1` (same for whatever equivalent value is in Render.com env)

akosasante: Right now, this line is creating a new socket connection every time we send a message. You can see in the video below that each time I send a message, a brand new websocket creation is created and in the backend each message seems to be coming from a new socket id:

One way to fix this by saving the socket in a variable in component state (please note, I'm not sure 100% if this is the best practice in React-land, but it will work well enough for this use case :sweat\_smile: ). So a `[socket, setSocket] = useState()` call. And then when the component first mounts in your `useEffect` on line 20, connect with `io(...)` as you do already, and do a `setSocket(...)` to save that socket to the component state.

Then here in the `handleSendMessage` you can use that `socket` variable instead of reconnecting with `io(...)` again.

src/components/album/AlbumChat.js
```diff
@@ -1,23 +1,32 @@
...
+    const regexPattern = /\/albums\/([a-zA-Z0-9]+)/;
+    const match = apiUrl.match(regexPattern);
```
akosasante: Instead of having to do this here and in `useEffect` we can just create a variable (doesn't need to be component state, just a regular `const` is fine here) at the top of the component that calculates the albumId or sets it to `null` or undefined otherwise.

akosasante (commented):
looking good; a few things to clean up and then the "sending"  should be  ready to merge

src/components/album/AlbumChat.js
mvstoyan: socket
this is what i have on backend now, when i sent message from /chat page input. User=undefined

akosasante (commented):
Looking great :+1: 

Only thing missing now, is that the `chat:album` socket endpoint expects an object with `message`, `albumId`, and `userId`. But we're only passing the first two.

  
In this react app, we are using the `react-auth-context` library that I've forked here: [https://github.com/akosasante/react-auth-context](https://github.com/akosasante/react-auth-context) to handle any auth, and keeping track of the currently logged-in user.

To get the currently logged-in user we can use this library's `useAuth` hook like this: `const { user: loggedInUser } = useAuth();` . The object inside `loggedInUser` is the full payload from the `/currentUser` endpoint in the backend. So to get the user id we'd need to do `loggedInUser.user.id` . Those can be done at the top of the component, and then pass that along in our request to the socket in `handleSendMessage`

mvstoyan:
socket
look like its work🎉

src/components/album/AlbumChat.js
```diff
@@ -1,40 +1,66 @@
...
-    if (match) {
-      // Extract albumId from the URL
-      const albumId = match[1];
-      // Сonnection is established with the Socket.io server
-      socket.on('connect', () => {
-        // Request to the server to join the album chat
-        console.log('Connected to WebSocket server');
-        socket.emit('join:album_chat', albumId);
+  const { user: loggedInUser } = useAuth();
+  const userId = loggedInUser ? loggedInUser.user.id : null;
```
akosasante: This works; we could also write 

```suggestion
  const userId = loggedInUser?.user?.id;
```

src/components/album/AlbumChat.js
```diff
@@ -1,40 +1,66 @@
...
+
+  // Nandle sending a chat message
```
akosasante:
```suggestion
  // Handle sending a chat message
```

akosasante (approved):
LGTM

PR #52, opened 2023-10-06
frontend for PS188

mvstoyan:
Some fixes for the PS188 backend.

### DIFF
--- a/src/components/album/AlbumChat.js
+++ b/src/components/album/AlbumChat.js
@@ -12,6 +12,8 @@ const AlbumChat = ({ apiUrl }) => {
   const { user: loggedInUser } = useAuth();
   const userId = loggedInUser?.user?.id;
 
+  const [messages, setMessages] = useState([]);
+
   // Handle sending a chat message
   const handleSendMessage = () => {
     if (socket && albumId) {
@@ -36,6 +38,9 @@ const AlbumChat = ({ apiUrl }) => {
         newSocket.emit('join:album_chat', albumId);
         newSocket.on('chat:album', (data) => {
           console.log('Received message from chat:album:', data);
+
+          // new message
+          setMessages((prevMessages) => [...prevMessages, data]);
         });
       });
       setSocket(newSocket);
@@ -53,6 +58,13 @@ const AlbumChat = ({ apiUrl }) => {
     <div>
       <h2>Album Chat</h2>
       <div>
+        <div>
+          {messages.map((message, index) => (
+            <div key={index}>
+              {message.userName}: {message.message}
+            </div>
+          ))}
+        </div>
         <input
           type="text"
           placeholder="Enter message"
--- a/src/components/album/AlbumChat.js
+++ b/src/components/album/AlbumChat.js
@@ -57,10 +57,12 @@ const AlbumChat = ({ apiUrl }) => {
           console.log('Received message from chat:album:', data);
 
           const isOwnMessage = loggedInUserId === data.user._id; // Determine isOwnMessage
-          setMessages((prevMessages) => [
-            ...prevMessages,
-            { ...data, isOwnMessage },
-          ]);
+          const messageWithKey = {
+            ...data,
+            isOwnMessage,
+            messageKey: uuidv4(),
+          }; // will add key to the message
+          setMessages((prevMessages) => [...prevMessages, messageWithKey]);
 
           scrollToBottom();
         });
@@ -103,7 +105,7 @@ const AlbumChat = ({ apiUrl }) => {
         ref={messagesContainerRef}
       >
         {messages.map((message) => (
-          <div key={uuidv4()}>
+          <div key={message.messageKey}>
             <div
               style={{
                 display: 'flex',
--- a/src/components/album/AlbumChat.js
+++ b/src/components/album/AlbumChat.js
@@ -11,10 +11,10 @@ const AlbumChat = ({ apiUrl }) => {
   const albumId = match ? match[1] : null;
 
   const { user: loggedInUser } = useAuth();
-  const userId = loggedInUser?.user?.id;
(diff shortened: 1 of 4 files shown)


src/components/album/AlbumChat.js
```diff
@@ -53,6 +58,13 @@ const AlbumChat = ({ apiUrl }) => {
     <div>
       <h2>Album Chat</h2>
       <div>
+        <div>
+          {messages.map((message, index) => (
+            <div key={index}>
```
akosasante: ⚠️  This is a common pitfall in React 😢 

The official React docs explain it better than me: https://react.dev/learn/rendering-lists

But basically, the issue is that every time we add a new message to the state (ie: with `setMessages`), React re-renders this component, that means the `index` that was there before is now going to be a new `index` (even if it ends up being the same number, it'll be a new "instance" of that number because React is going to do `messages.map` again). 

Instead, React wants us to use a value that it can keep track of across renders. 

In this case, we can handle this in the frontend or the backend (or a mix of the two):
**Handle in the frontend**
For the messages we get from the `'chat:album'` socket listener, we can add an "key" property to the `data` payload, with the official doc's suggestion of `crypto.randomUUID()` to generate the key. This can then be used as the key here when we iterate over them

**Handle in the backend**
If the socket handler in the backend is going to save `ChatMessage` to the database anyway when it receives them, then it can return that mongo object back when it emits a message; that will have a mongo `_id` field that we know is unique and we safely use that field as the iteration key here in the React code.

akosasante (commented):
Looking great! just one fix needed and you should be good to merge

src/components/album/AlbumChat.js
akosasante: Since in this case, there's already work being done by you on the backend to save the ChatMessage when receiving these messages, then that is likely the best road to go down. But wanted to present both options just for learning purposes that this can be done even if the object you're iterating over is not something that is being saved in the db.

mvstoyan: I made a little temporary design just to see how it would turn out. But later, we will redo the entire design when we decide exactly how it will look full app.

mvstoyan: On the backend, I'm receiving a response like the one in the screenshot. However, on the frontend, I'm trying to display the username:message properly instead of :message. But not work for me

src/components/album/AlbumChat.js
```diff
@@ -10,15 +11,19 @@ const AlbumChat = ({ apiUrl }) => {
...
 
   // Handle sending a chat message
   const handleSendMessage = () => {
     if (socket && albumId) {
       const messageData = {
         message,
         albumId,
-        userId,
+        user,
+        isOwnMessage: true,
```
akosasante: Currently, when userA **_sends_** a message, they're sending it to the socket with `isOwnMessage: true`. And when userB (and all the other users in the same chat room) _**receives**_ that message on their browser, it will still say `isOwnMessage: true` because the backend was told to just forward along the `data` it was given. 

A different approach I would suggest, is _**not**_ including `isOwnMessage` in the message we send to the backend. Instead, when we receive the message in the frontend (in the `chat:album` listener on line 56) we can compare the id of the user who sent that message, to the logged-in user id that this component has on line 14 from the `useAuth` hook and set the `isOwnMessage` value there.

Here's some pseudocode to illustrate my suggestion:

```
newSocket.on('chat:album', (data) => {
   ...existing code...

  const isOwnMessage = loggedInUserId === data.user._id;

  setMessages((prevMessages => [...prevMessages,  { ...data, isOwnMessage } ]);
}
```

src/components/album/AlbumChat.js
```diff
@@ -49,17 +81,87 @@ const AlbumChat = ({ apiUrl }) => {
...
+                style={{
+                  backgroundColor: message.isOwnMessage
+                    ? '#3f51b5'
+                    : 'lightblue',
+                  color: message.isOwnMessage ? 'white' : 'black',
+                  borderRadius: '8px',
+                  padding: '8px 16px',
+                }}
+              >
+                <strong>{message.user.username}</strong>: {message.message}
```
akosasante: Your log line on line 57 is showing me that the messages that are being returned by the backend do not have a `message.user.username` field. I've left review comments in the backend to address that

src/components/album/AlbumChat.js
akosasante: This initial design looks fantastic Marina, great job with it!

src/components/album/AlbumChat.js
```diff
@@ -47,19 +74,89 @@ const AlbumChat = ({ apiUrl }) => {
...
+          flex: 1,
+          padding: '16px',
+          width: '95%',
+          maxHeight: '65vh',
+          overflowY: 'auto',
+        }}
+        ref={messagesContainerRef}
+      >
+        {messages.map((message) => (
+          <div key={uuidv4()}>
```
akosasante: [https://react.dev/learn/rendering-lists](https://react.dev/learn/rendering-lists)

> Similarly, do not generate keys on the fly, e.g. with `key={Math.random()}`. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.

If you want to generate an id here in the frontend (rather than returning the `ChatMessage` object from the backend which will have an `_id` field), that's totally fine :+1: but we should not generate the ID within the rendered JSX based on the above info from the offical React docs.

Instead, we should generate them in the component outside of the `return (...)` part so that React can keep track of the value between renders. My suggestion would then be to do so when we call `setMessages` on line 60. We can do something like `data.messageKey = uuidv4()`. Or you can use object destructuring to do it on one line if you prefer.

akosasante (approved):
Fantastic job Marina! :ship: :rocket:

PR #53, opened 2023-10-22
Fixed warnings with tooltips and buttons

AlonaVladymyrovaTrinity:
Fixed warnings with tooltips and buttons in files:
src/components/layout/AddToCartButton/AddToCartButton.js
src/components/layout/AddToWishlistButton/AddToWishlistButton.js


akosasante (approved):
Thanks! :rocket:

PR #55, opened 2023-10-30
PS155-PropTypes-Album/AlbumGrid

mvstoyan:


src/components/album/Album.js
```diff
@@ -75,4 +76,11 @@ function Album({ album, wishListId }) {
   );
 }
 
+Album.propTypes = {
+  album: PropTypes.shape({
+    _id: PropTypes.string.isRequired,
+  }).isRequired,
+  wishListId: PropTypes.string,
+};
```
akosasante: This is good; While it's not necessarily needed to put in _all_ possible fields of the album object;  I would recommend at least putting in the types of all the fields that are used in this file.

akosasante: We can further validate using custom [PropType validation,](https://designcode.io/react-hooks-handbook-custom-proptypes) that for example `album.releaseDate` is a string in the expected date format, or that `averageRating` is a number between 1 and 5 as we would expect

src/components/album/AlbumGrid.js
```diff
@@ -14,4 +15,13 @@ function AlbumGrid({ albums, wishListId }) {
   );
 }
 
+AlbumGrid.propTypes = {
+  albums: PropTypes.arrayOf(
+    PropTypes.shape({
+      _id: PropTypes.string.isRequired,
+    })
```
akosasante: If you want to avoid having to retype long proptypes like the full album shape in multiple places, you can also centralize them.

You could have a folder `src/propTypes` with files for different prop-types eg: `albumTypes.js` and from there, do an `export const albumShape = PropTypes.shape(...)`. Then in both `Album.js` component and in `AlbumGrid.js` component you can import and reuse that prop type within the code (eg: `albums: PropTypes.arrayOf(albumtype)` )

akosasante (approved):
LGTM, left some suggestions for additional thoughts around document component types with prop-types

mvstoyan:
@akosasante Can you check please, I'm not sure if it's correct

src/propTypes/albumTypes.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+
+export const albumShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  albumName: PropTypes.string.isRequired,
+  artistName: PropTypes.string.isRequired,
+  image: PropTypes.string.isRequired,
+  // custom validator function takes three arguments
+  releaseDate: (props, propName, componentName) => {
+    if (!/^\d{4}-\d{2}-\d{2}$/.test(props[propName])) {
```
akosasante: Currently, the backend returns a `releaseDate` that is a full date + time timestamp:

Screenshot 2023-11-03 at 9 02 13 AM

Your regex is currently matching on just a date (4 digits, a dash, 2 digits, a dash, 2 digits, and then the end of the string).

You can use https://regex101.com/ to play around with regex to find the correct regex to match the expected strings:

Screenshot 2023-11-03 at 9 03 40 AM

src/propTypes/albumTypes.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+  },
+  averageRating: (props, propName, componentName) => {
+    if (
+      typeof props[propName] !== 'number' ||
+      props[propName] < 1 ||
+      props[propName] > 5
+    ) {
+      return new Error(
+        `Invalid prop ${propName} supplied to ${componentName}. Validation failed.`
```
akosasante: Not a big deal, but for both of these custom props it might be nice to have a bit more info in the error returned, about _why_ validation failed. For example for the rating, we could say "Validation failed. averageRating must be a number between 0 and 5". 

Makes it a bit easier for another developer to know what's gone wrong, without having to go and open up this file to inspect the exact validation code.

src/propTypes/albumTypes.js
```diff
@@ -0,0 +1,28 @@
...
+      // format YYYY-MM-DD
+      return new Error(
+        `Invalid prop ${propName} supplied to ${componentName}. Validation failed.`
+      );
+    }
+  },
+  averageRating: (props, propName, componentName) => {
+    if (
+      typeof props[propName] !== 'number' ||
+      props[propName] < 1 ||
```
akosasante: According to where our model is defined in the backend, it is possible for averageRating to be 0 if no one has reviewed this album yet: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/models/Album.js#L50

src/propTypes/albumTypes.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const albumShape = PropTypes.shape({
+  _id: PropTypes.string.isRequired,
+  albumName: PropTypes.string.isRequired,
+  artistName: PropTypes.string.isRequired,
+  image: PropTypes.string.isRequired,
+  // custom validator function takes three arguments
+  releaseDate: (props, propName, componentName) => {
+    if (!/^\d{4}-\d{2}-\d{2}$/.test(props[propName])) {
+      // format YYYY-MM-DD
```
akosasante:
```suggestion
YYYY-MM-DD[T]HH:MM:SS.MS[Z] # formatted as a date-time string in UTC timezone.
```

src/propTypes/albumTypes.js
```diff
@@ -0,0 +1,28 @@
...
+      return new Error(
+        `Invalid prop ${propName} supplied to ${componentName}. Validation failed.`
+      );
+    }
+  },
```
akosasante: Out backend [model](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/models/Album.js#L30) says this field is not required. So it's possible sometimes it will be null/undefined. Currently, this code means it would fail validation in that case (because the regex wouldn't match). But we probably just want to skip checking the regex in that case and treat the value as allowed/valid.

```suggestion
releaseDate: (props, propName, componentName) => {
    if (props[propName]) {
      if (!/^\d{4}-\d{2}-\d{2}$/.test(props[propName])) {
        // format YYYY-MM-DD
        return new Error(
          `Invalid prop ${propName} supplied to ${componentName}. Validation failed.`
        );
      }
    }
  },
```

akosasante (commented):
Looking good; left some notes. 

For proptypes, since these are something that are checked at "run time" (meaning instead of running a lint command, we actually only see the errors by running the code), I like to test it by manually entering something into the props that I know won't pass. For example here I am passing an object that only has id and album name, into the `<Album>` component:

```js
function AlbumGrid({ albums, wishListId }) {
  const myAlbum = {_id: "1", albumName: "bro"};
  return (
    <Grid container spacing={2}>
      {albums.map((album) => (
        <Grid item key={album._id} xs={12} sm={6} md={4} lg={3}>
          <Album album={myAlbum} wishListId={wishListId} />
        </Grid>
      ))}
    </Grid>
  );
}
```

When I run the app locally, I can now see the validation errors in the dev console, telling me that the first validation step, for `artistName` is failing.

Screenshot 2023-11-03 at 9 20 43 AM

I then continue to test manually by entering different valid or invalid values and checking that the expected console message (or no console message) is shown.

src/propTypes/albumTypes.js
mvstoyan: screen

src/propTypes/albumTypes.js
mvstoyan: @akosasante Can you check, if it's correct. Look like work😅

akosasante (approved):
LGTM 🚢

src/propTypes/albumTypes.js
akosasante: 👍

PR #56, opened 2023-10-30
upd user info without stying

NatachaKey:


src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
+import React, { useEffect, useState } from 'react';
+import { Card, CardContent, Table, TableRow, TableCell, TextField, Button } from '@mui/material';
+import { Avatar } from '@mui/material';
+import axiosInstance from '../../apis/axiosClient';
+
+export default function PersonalProfile() {
```
akosasante: It doesn't actually matter too much but for clarity, let's call this component something like `PersonalProfileEditForm` or something like that to differentiate it from the component with the same name in `UserProfile.js`

src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
...
+  const [successMessagePassword, setSuccessMessagePassword] = useState('');
+
+  const fetchCurrentUser = async () => {
+    try {
+      const response = await axiosInstance.get('/users/showMe');
+      setUserData(response.data.user);
+    } catch (error) {
+      console.error('Error fetching current user', error);
+    }
+  };
```
akosasante: Since the user is already logged in; we can actually get this data from the `useAuth()` hook provided by the auth library that we're using in the frontend, [react-auth-context](https://github.com/akosasante/react-auth-context).

One thing the library does is checks if the user still has valid cookies upon each route navigation. It does this by calling the `/users/showMe` endpoint. It stores the returned user in a [Context State](https://react.dev/reference/react/createContext) variable; which we can retrieve on any page by doing: `{ user } = useAuth()`. [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/76b7ae6fbe5d5660fb0915025bc3e7b51bb83dae/src/components/album/AlbumChatBox.js#L14](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/76b7ae6fbe5d5660fb0915025bc3e7b51bb83dae/src/components/album/AlbumChatBox.js#L14)

This allows us to avoid making duplicate calls to the backend for data already available in the frontend

One messy thing is that the user that's returned is nested in its original response shape so to use the actual user you'll need to unwrap it for now until we clean that up (`const userData = user.user` or `const { user: { user: userData }} = useAuth();`)

src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
...
+          />
+        </div>
+        {successMessage && (
+          <div className="text-success">{successMessage}</div>
+        )}
+        {successMessagePassword && (
+          <div className="text-success" >{successMessagePassword}</div>
+        )}
+        <form onSubmit={handleProfileSubmit}>
+          <Table responsive striped hover className="text-center mt-5">
```
akosasante: There's errors in the console due to these fields. I think when the `UserProfile.js` component was made that used these fields, we were planning to go back to fix that.

The errors are due to these fields not being valid for material-ui (I think when the UserProfile component was used there was some confusion between the syntax for styling material-ui table vs bootstrap-based table).



```suggestion
          <Table className="text-center mt-5">
```

Same for the `<Table>` component on line 164

src/components/user/UpdateUserProfileFields.js
akosasante, line 1: After saving, there should be a way to navigate back to my user profile page. Or perhaps it should auto-navigate me upon a successful save.

src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
...
+              maxHeight: '100px',
+            }}
+          />
+        </div>
+        {successMessage && (
+          <div className="text-success">{successMessage}</div>
+        )}
+        {successMessagePassword && (
+          <div className="text-success" >{successMessagePassword}</div>
+        )}
```
akosasante: These look good; you might also consider Snackbar/Toast as a type of component for making success/error messaging clear and eye-catching: https://mui.com/material-ui/react-snackbar/

src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
...
+              <TableRow>
+                <TableCell>EMAIL</TableCell>
+                <TableCell>
+                  <TextField
+                    name="email"
+                    value={formData.email}
+                    onChange={handleInputChange}
+                    error={errors.email ? true : false}
+                    helperText={errors.email}
+                  />
```
akosasante: `TextField` material-ui component allows us to give it a `type` attribute that is passed to the underlying `input` element: [https://mui.com/material-ui/react-text-field/](https://mui.com/material-ui/react-text-field/)

[https://www.w3schools.com/html/html_form_input_types.asp](https://www.w3schools.com/html/html_form_input_types.asp)

This can give us some browser-built in validation. For example if I make this field a `type="email"` and try to type in a not-email , the browser will show a warning, without us having to write any custom validator code or regex or anything:

src/components/user/UpdateUserProfileFields.js
```diff
@@ -0,0 +1,201 @@
...
+        setSuccessMessage('User profile saved successfully.'); // Set success message
+        setFormData({ // Clear the form
+          name: '',
+          email: '',
+          oldPassword: '',
+          newPassword: '',
+        });
+      } catch (error) {
+        // Handle errors and update the state with validation errors
+        setErrors(error.response.data.errors);
```
akosasante: The error code in the rest of this component assumes that the error will always be an object. Here we are ending up setting `errors` to `undefined`. Which then results in errors breaking the page. I tested this by submitting an update where I passed in a non-email as my email. This causes the backend to return a 400, which axios will raise as an error.



When I log `error` variable to the page I get:



So the backend is returning the error as an object with the field `msg` (which we could get with `error.response.data.msg`) but we are trying in the code to access `error.response.data.errors` which is not an available field and will return `undefined`. 

---

A problem you'll run into after fixing this is where to put this error in your error object. Since without reading the text of the error, we won't know if it's about the email or the name or something else completely. 

One approach could be to have another field on your error object for `responseError` or something. And show that in a different way on the page, maybe using a Snackbar component or red text at the top of the form rather than under a specific input box.

akosasante: Same for the `catch` clause on `handlePasswordSubmit`

src/components/user/UpdateUserProfileFields.js
NatachaKey: Thank you @akosasante , done!

src/components/user/UpdateUserProfileFields.js
NatachaKey: @akosasante thank you!! totally forgot about it

src/components/user/PersonalProfileEditForm.js
```diff
@@ -14,20 +15,13 @@ export default function PersonalProfile() {
...
-      console.error('Error fetching current user', error);
-    }
-  };
-
-  useEffect(() => {
-    fetchCurrentUser();
-  }, []);
-
+  const { user } = useAuth(); 
+   let currentUserData = user.user ; //he user that's returned is nested in its original response shape so to use the actual user you'll need to unwrap it
```
akosasante: This variable is called `currentUserData` but I'm not seeing it used anywhere in the component. In the component we refer to `userData`. We can remove line 8, and rename this variable to `userData` since we no longer need to manually set user data ourselves, since we're just getting the current value from the `useAuth` hook.

The linter message you've mentioned in the comment is saying that you've marked this variable as mutable (`let`) but it's never being mutated/it's value is never being changed in the component, so it it should be `const` instead.

src/components/user/PersonalProfileEditForm.js
```diff
@@ -58,9 +52,12 @@ export default function PersonalProfile() {
           oldPassword: '',
           newPassword: '',
         });
+        // Navigate back to the user profile page
+        navigate('/profile'); // Use the navigate function
       } catch (error) {
         // Handle errors and update the state with validation errors
-        setErrors(error.response.data.errors);
+        setErrors(error.response.data.msg);
```
akosasante: We're still prone to some issues with the two `setErrors` calls unfortunately.

`error.response.data.msg` returned by the backend is a string that explains the error (shown in screenshot here: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/56#discussion_r1377507040) 

But in our code here in this component, we are expecting the `errors` variable to always be an object (for example on line 141 we have code that says `error={errors.name ? true : false}`. If we've set `errors` here to the string "Please provide valid email", then `errors.name` is going to fail and raise an error on the page.

Unless we want to change the return value in the backend (I think probably not), we should be handling this here on the client-side. 

Since we don't really know what the `msg` is saying without doing some regex or something, we don't know if the error message is about the email or about the name (or even just a general 500 "Internal Error" message from the server). 

So instead one approach would be to put this msg into another field on the `errors` object. Eg: `setErrors({ serverMsg: error?.response?.data?.msg })`. Also note how I used optional chaining `?.` because it's possible that the error doesn't have this exact shape (if the error happens within our handle input code itself or if the server returns something else or if the internet is disconnected and we get a "failed to make http call" error from the browser). So to be safe, I like to use optional chaining (or if you prefer you can do if statements).

Then we can show that error in the `Alert` per below comment. If you prefer to not extend the existing `errors` variable since currently that's mainly being used for errors about the form input itself; you could instead have a separate `serverErrors` state variable that you set here (and make sure to clear when there is a successful request/new request being made).

src/components/user/PersonalProfileEditForm.js
```diff
@@ -0,0 +1,205 @@
...
+        setFormData({ // Clear the form
+          name: '',
+          email: '',
+          oldPassword: '',
+          newPassword: '',
+        });
+      } catch (error) {
+        setErrors(error.response.data.msg); 
+        // Handle errors and update the state with validation errors
+        <Alert severity="error">{errors}</Alert>
```
akosasante: Putting an element in here won't render on the page. Only things in `return (...)` will be rendered.

akosasante: Instead, I would move this error alert to be in the card content. Maybe right under the "My Profile" header. 

```
{errors.serverMsg && <Alert severity="error">{errors.msg}</Alert>}
```
^ This will make React only render the error message _if_ `errors.serverMsg` (or whatever you choose to label errors from those catch clauses) is populated

NatachaKey: @akosasante Akos, could you please take a look- might be doing smth wrong- I don't see custom error msg from Snackbar

akosasante: I'm not sure what you mean; the `Alert` component seems to be correctly showing the error message from the server now from what I can see:

Screenshot 2023-11-03 at 6 08 02 PM

akosasante (approved):
Everything looks good to me ✅ 

But please let me know/clarify the issue you were seeing?

src/components/user/PersonalProfileEditForm.js
NatachaKey: @akosasante Akos thank you!! everything works correctly

akosasante:
@NatachaKey  it looks like this was merged with some failing github actions: https://github.com/Code-the-Dream-School/dd-prac-team4-front/actions/runs/6787780992/job/18451360086?pr=56

Please check if your `husky` library is set up as needed in the frontend the same way we resolved in the backend repo; to prevent this from reoccurring. I've pushed a format/lint fix to main for these.

PR #57, opened 2023-10-31
PS84-Redux-move-album-to-wishlist

mvstoyan:


mvstoyan:
console
Akos, the album was successfully added to the wishlist, but when requesting the wishlist data, the 'albums' array is returned as empty: []. Could you guide me if this indicates a problem on the server, or if I might be doing something wrong on the frontend?

src/components/OrderSidebar.js
```diff
@@ -23,6 +23,11 @@ const OrderSidebar = () => {
   const tax = useSelector((state) => state.cart.tax);
   const total = roundedNumber(useSelector((state) => state.cart.totalAmount)); //total amount after tax
 
+  const handleMoveToWishlist = (album) => {
+    dispatch(reduceItem({ album, quantity: 1 }));
```
akosasante: This should probably remove the item altogether from the order sidebar/cart so we should reduce it by the album's current quantity in the cart; at least the way I'm picturing it?

src/redux/shoppingCart.js
```diff
@@ -69,8 +70,21 @@ const shoppingCartReducer = createSlice({
 
       saveCartToLocalStorage(state);
     },
+    addToWishlist(state, action) {
+      const { album } = action.payload;
+      console.log('Album ID to add to wishlist:', album.id);
+      axiosInstance
+        .post('/wishlist/', { albumId: album.id })
```
akosasante: The `POST /api/v1/wishlist` endpoint that you're hitting here is a sort of get-or-create endpoint: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/926a305b7f6be72658f74b652b10b02cb33696b2/src/controllers/wishlistController.js#L6](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/926a305b7f6be72658f74b652b10b02cb33696b2/src/controllers/wishlistController.js#L6)

If the user doesn't already have a wishlist, it creates a new empty one. If they already had a wishlist, it returns that wishlist in the response. 

So we do want to make this call, but we can't pass the `albumId` here. Instead, we then must make a second call to add the album to the wishlist using the add album to wishlist endpoint: [https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/routes/wishlistRoutes.js#L23](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/1e48ca722a1a852752cda040cae126beeffd535d/src/routes/wishlistRoutes.js#L23)

A future improvement we could make is to have the user's wishlist be returned as part of the user info as soon as they are logged in or somewhere else earlier/higher up in the component chain, so that here we would already have the wishlist id and could go straight to adding the album. But for now, doing it this way is totally reasonable :smile:

mvstoyan: @akosasante  I wrote the logic in handleMoveToWishlist in OrderSidebar.js instead of shoppingCart.js because Redux doesn't allow directly using asynchronous functions. However, it seems we can achieve this through the Redux Toolkit library. I'm not very familiar with it, but if it's the right solution, I can give it a try. Or am I reasoning incorrectly?

src/components/OrderSidebar.js
```diff
@@ -78,12 +89,15 @@ const OrderSidebar = () => {
                   secondary={`Artist: ${item.album.artistName}, Price: $${item.album.price}`}
                 />
               </ListItem>
-              <ListItem>
+              <div>
```
akosasante, line 92: Any particular issue that led to this element change?

akosasante (approved):
LGTM 👍 

Good catch! I'll write up some tickets to have us set up a more consistent use of redux vs redux toolkit vs rtk query for async things. But this approach works well here ✅

mvstoyan:
> LGTM 👍
> 
> Good catch! I'll write up some tickets to have us set up a more consistent use of redux vs redux toolkit vs rtk query for async things. But this approach works well here ✅

I would love to learn this, thank you!

src/components/OrderSidebar.js
mvstoyan: No, this is just now, later we will move to the cart page and add style, we will remove this div. Now the button is on the text, without div

PR #58, opened 2023-11-03
changed the way to get users info/ useAuth

NatachaKey:


src/components/user/UserProfile.js
```diff
@@ -1,26 +1,26 @@
...
+  //     console.error('Error fetching current user', error);
+  //   }
+  // };
 
-  useEffect(() => {
-    fetchCurrentUser();
-  }, []);
+  // useEffect(() => {
+  //   fetchCurrentUser();
+  // }, []);
```
akosasante: We can remove this now altogether :)

akosasante (approved):
🚢

src/components/user/UserProfile.js
NatachaKey: done=)

PR #59, opened 2023-11-03
Display reviews bug

NatachaKey:


src/components/album/AlbumPreview.js
```diff
@@ -26,9 +28,24 @@ const AlbumPreview = ({
     }
     //extract the album id
     const albumId = match[1];
```
akosasante: Here we are extracting the Spotify Album ID (ie the identifier used by the Spotify API) from the url; because that's what the spotify preview component needed in order to render the "play clips from this album" stuff:

Screenshot 2023-11-03 at 5 40 54 PM

So in the above screenshot, we were taking the URL `"https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy"` and grabbing the part `6r1lh7fHMB499vGKtIyJLy` and calling that `albumId`. If it helps make it clearer for future use of this component, we can rename that variable to `spotifyAlbumId`.

In your use case of rendering the reviews, you need the mongo album ID as you've said. Well, the parent component that is rendering this `AlbumPreview` has that information, so we can update this component's props to accept the full album instead of just the spotify URL, or if we want, we can have the parent pass in the album id if that's the only part of the album object that we need here. No need to make a call to the backend, when we have this information.

--- 

One thing to note, is that `<AlbumPreview>` is being used in two places right now: 
1. In the `Album` component: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/76b7ae6fbe5d5660fb0915025bc3e7b51bb83dae/src/components/album/Album.js#L71
2. In the `AlbumChat` component (eg: `localhost:3000/chat` where users can chat in real-time with each other about an album they have purchased): https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/76b7ae6fbe5d5660fb0915025bc3e7b51bb83dae/src/components/album/AlbumChat.js#L1-L10. Now in this case, I'm not sure that we want to also be showing reviews, it might make the page a bit cluttered

One way we can handle this is by having some conditional code inside the component (eg: if "albumId" was passed in then show reviews. or if "showReviews" is true then show the reviews component).

Another way, that may be more preferable in this case, is that all React components have a special prop called `children`: https://react.dev/learn/passing-props-to-a-component#passing-jsx-as-children

We are already using it here so that the `AlbumChat` component can be rendered as a _child_ of this `<AlbumPreview>` component on line 63.

We could do the same with `<ReviewsList>` which would mean we don't need to change how this component works too much and have logic about reviews here where it should mostly be concerned with logic about "previewing albums".

In `Album.js` we can have code like:
```jsx
          <AlbumPreview spotifyUrl={album.spotifyUrl}>
            <AlbumReviews albumId={album._id} />
          </AlbumPreview>
```

By nesting `<AlbumReviews>` inside of `<AlbumPreview>`, this will cause the review component to be passed to the AlbumPreview component as the `children` prop.

You may need to add new props to `<AlbumPreview>` to maybe handle the styling/positioning of the children but that should at least get you the two things rendered together. And since Album.js already has fetched the album info, you only need to pass it into `AlbumReviews` itself, rather than having to pass it through `AlbumPreview` which doesn't need that info for its own logic.

NatachaKey: @akosasante Akos, just added the reviews preview, no idea how to style it correctly- how to change it to  make it appear beow the album card?

akosasante: I left some comments in your Write Review PR that I think cover this question. :+1: :slightly\_smiling\_face:

NatachaKey: @akosasante thank you!!

PR #60, opened 2023-11-03
Theme switcher

DYA13:
Akos , I would like to ask if I need to pass as props in component Home ? As at the moment in console ThemeSwitcher.js seems to be working but practically doesn't change the theme for dark.


src/components/Navbar.js
```diff
@@ -17,8 +17,9 @@ import MenuIcon from '@mui/icons-material/Menu';
 import './Navbar.css';
 
 import Logo from '../images/Logo.png';
+import ThemeSwitcher from './ThemeSwitcher';
 
-const Navbar = () => {
+const Navbar = ({ isDarkTheme }) => {
```
akosasante: Nothing is currently passing this prop into `Navbar` when it's used, so currently this prop variable, `isDarkTheme` will always be undefined.

We can inspect this using the React Dev Tools extension: https://react.dev/learn/react-developer-tools
Screenshot 2023-11-03 at 5 31 08 PM

Or by logging the value when the component loads:
```js
const Navbar = ({ isDarkTheme }) => {
  console.log("isDarkTheme prop: ", isDarkTheme);
//...etc
```
you'll see that this logs just once when the component is loaded, and is not reset when we click the theme switcher button:
Screenshot 2023-11-03 at 5 30 40 PM

---

In the `ThemeSwitcher` component we are changing a _state variable_ `isDarkTheme` with `useState`. Keep in mind that state variables are _local to the component_. The parent component (Navbar) won't automatically know about state variables inside its children. In React, we usually say that "state flows down via props". That is to say, we can take a state variable and let a _child component_ know about it by passing it in via props. 

If we want to tell parent components about a change in a variable we can use callback functions or we can use React Context: https://react.dev/learn/sharing-state-between-components, https://react.dev/learn/passing-data-deeply-with-context. Or we can make state global using Redux. 

--- 

Overall, in this case, since we are already using the `material-ui` library to handle the styling of our pages; I would recommend trying to hook into that. That way, every component that wants to have a dark style, doesn't need to do it piecemeal by checking for a state variable and then changing individual style attributes.

Here is an example from the Material-UI docs: https://mui.com/material-ui/customization/dark-mode/#toggling-color-mode

here is a bit more concise example someone has on CodeSandbox: https://codesandbox.io/s/mui-react-theme-switcher-0reg4?file=/src/App.tsx

Hopefully that can get you started, but feel free to let me know if you have any questions!

DYA13: Akos , I have taken ui material code as an example to change color of the theme but I am stuck a bit how to pass it throughout all components

akosasante: `ThemeProvider` component is a React Context Component: https://react.dev/reference/react/createContext

These are special types of components that let us essentially create global state that any component that is a child of that Context Component can have access to.

We already use Context Components in our application for a few use cases. Generally the common practice is to set up Context Components at the very top/root of your application. So in our application we have these set up in `index.js`:

```jsx
  <ThemeProvider theme={defaultTheme}>
    {/* Use CSS baseline to set common spacing/sizing across all of our styling */}
    <CssBaseline />
    <BrowserRouter>
      <AuthProvider {...authSettings}>
        <Provider store={store}>
          <App />
        </Provider>
      </AuthProvider>
    </BrowserRouter>
  </ThemeProvider>
```

- `BrowserRouter` component gives all the child components access to `react-router`'s global state and hooks like `useNavigation` for linking to different routes we've defined in `App.js`
- `AuthProvider` component gives all the child components access to `react-auth-context`'s global state and hooks like `useAuth` for checking the logged in user, submitting login/logout forms, etc
- `Provider` component gives all the child components of the app access to `redux` global state and the ability to dispatch actions and fetch state from the Redux store

In this PR, we are interested in the `<ThemeProvider>` component: https://mui.com/material-ui/customization/theming/#:~:text=Theme%20provider,API%20section. The docs in that link tell us that:

> `ThemeProvider` relies on the [context feature of React](https://react.dev/learn/passing-data-deeply-with-context) to pass the theme down to the components, so you need to make sure that `ThemeProvider` is a parent of the components you are trying to customize. You can learn more about this in [the API section](https://mui.com/material-ui/customization/theming/#themeprovider).

So **we should not** need to add `<ThemeProvider>` anywhere else in our application since it's already been added at the very top of the application. Everything inside the app should have access to the theme.

The default theme is also defined in `index.js` and passed into `ThemeProvider` to initialize it: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/76b7ae6fbe5d5660fb0915025bc3e7b51bb83dae/src/index.js#L29-L32

You'll see that if you change, the value for the `primary.main` color in there for example, then all the buttons on the page which use that value will change their color:

Screenshot 2023-11-10 at 10 43 54 AM


Screenshot 2023-11-10 at 10 42 23 AM

---

In terms of toggling, we can see in the docs here that changing the `mode` to `"dark"` will change the overall theme of the page: https://mui.com/material-ui/customization/dark-mode/#dark-mode-by-default:

Screenshot 2023-11-10 at 10 45 02 AM

Screenshot 2023-11-10 at 10 45 08 AM

In order to make this configurable, the example in Material-UI uses the approach of creating a custom React Context Component that they have named `ColorModeContext` and setting that Provider as the _parent_ of the theme Context Component. That's a fine approach but requires some refactoring to work in our app. We would need to move the `<ThemeProvider>` component from being defined in plain javascript in the `index.js` file to instead being defined within the `App` component in `App.js`. Then you can move a bunch of the code you've written in `ThemeSwitcher.js` based on the code in the material-ui docs, to instead be in `App.js`.

---

Alternatively, I find that the approach in the second example given in the earlier comment will be a bit more straightforward to implement into our existing application: https://codesandbox.io/s/mui-react-theme-switcher-0reg4?file=/src/App.tsx:1046-1149

In that one, instead of creating a custom Context for the color mode, they are creating a function `toggleDarkMode` and passing that down to their theme switching button component. That function lives in the `App` component where they are defining their `ThemeProvider`. So in this approach, we do still need to move `<ThemeProvider>` (and the associated `<CssBaseline>` which standardizes our styling) down to `App.js` but we don't need to create any new Context components in this approach.

Often in React you will see that there are multiple ways to get state down from a component higher up in the DOM tree to a component lower down. We can pass down functions or state via props. We can use React Context Components. We can use the global Redux store. In this case we will go with passing down via props since we're only needing to pass down the value through one or two components.

In `index.js`, we can remove `<ThemeProvider>`, `<CssBaseline>` and `defaultTheme`, and move those to `App.js` instead

In `App.js` we should then have:
- `useState` function for getting and setting the current color/theme mode
- `createTheme` from `@mui/material` dependency that sets up the initial theme, and passes in the `mode` variable (this is **instead of** having a static `defaultTheme` variable like we did in `index.js`)
- `toggleDarkMode` function that takes a value or boolean, and updates the color mode using the set state function from bullet 1
- In the return value of the component, `<ThemeProvider theme={theme}>` should be the parent element of everything else
- `Navbar` component should accept a prop where we pass in the `toggleDarkMode` function

In `Navbar.js` we just want to pass that `toggleDarkMode` function down into our `ThemeSwitcher` component

In your `ThemeSwitcher` component you can return to the code you had in commit `3a3334dc2d3cdcbe58c7b1e3c2ec9e1db98d68ba` (https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/3a3334dc2d3cdcbe58c7b1e3c2ec9e1db98d68ba/src/components/ThemeSwitcher.js(

You can use the Theme Context hook function `useTheme()` to get the current value of the theme (eg to see if it's currently in dark mode or light mode: `theme.palette.mode`)

When the user clicks the theme switcher button, we can call the `toggleDarkMode` function, which will change the state of the variable `mode` all the way up back in `App.js` for us.

DYA13: Thank you very much for such a detailed explanation❤️ I appreciate it a lot. I have done the refactoring of the code as you suggested, and it works now. I just wanted to ask about ThemeSwitcher.js, specifically lines 11-21. I tried to play with the palette, the primary and secondary colors, but I don't see the application of these colors when toggling the button. Maybe I forgot to pass them somewhere in the components🤔

const modifiedDarkTheme = createTheme(theme, {
    palette: {
      mode: 'dark',
      primary: {
        main: '#2196f3',
      },
      secondary: {
        main: '#ff4081',
      },
    },
  });

src/components/ThemeSwitcher.js
```diff
@@ -0,0 +1,44 @@
...
+        main: '#2196f3',
+      },
+      secondary: {
+        main: '#ff4081',
+      },
+    },
+  });
+
+  return (
+    <ThemeProvider theme={modifiedDarkTheme}>
```
akosasante, line 24: This `ThemeProvider` component and the `modifiedDarkTheme` that you're passing it; will only affect the _children_ of this component. So in this case, the `primary`, and `secondary` palette values that you've chosen will only affect the `Box`, `IconButton`, and BrightnessIcon elements you have here; but nothing elsewhere in the app or page

akosasante: If we want to change the `primary.main` and `secondary.main` values of the overall theme; we need to add conditional values up in `App.js` where the `ThemeProvider` component is wrapping the _entire application_. 

Right now, when `toggleDarkMode` is called here in this component, it will cause the `mode` state variable in `App.js` to be changed. 

What we can do then is if you want to completely customize your themes based on the light/dark mode, is something like:

```jsx
// in App.js
  const [mode, setMode] = useState('light');

  const toggleDarkMode = (value) => {
    setMode(value);
  };

  const lightTheme = createTheme({
    palette: {
      mode: 'light',
      primary: { main: '#3f51b5' }
      // etc any other customizations that we want to make for light mode
    }
  });

  const darkTheme = createTheme({
    palette: {
      mode: 'dark',
      primary: { main: '#2196f3' }
      // etc any other customizations that you want to make for dark mode
    }
  });

  // set the currentTheme variable based on the `mode` state variable which may be toggled with the `toggleDarkMode` function
  const currentTheme =  mode === 'dark' ? darkTheme : lightTheme;

    return (
      <ThemeProvider theme={theme}>
      // etc
    );
}

DYA13: Thank you Akos , I have implemented your code 👍

akosasante (approved):
Great job!

PR #61, opened 2023-11-05
fetchRecentMessages

mvstoyan:


mvstoyan:
I still need to replace the user ID with the name. But I want to understand if I'm doing it right at the moment😅

src/components/album/AlbumChatBox.js
```diff
@@ -47,28 +47,44 @@ const AlbumChatBox = ({ spotifyUrl }) => {
   };
 
   useEffect(() => {
+    const generateMessageWithKey = (data) => {
+      const isOwnMessage = loggedInUserId === data.user._id;
+      return {
+        ...data,
+        isOwnMessage,
+        messageKey: uuidv4(),
```
akosasante: I don't believe that we still need to create our own `messageKey` here in the frontend? Now we are returning the full `ChatMessage` model from the backend which already has a unique key that we can tell React to use when rendering a list of messages (`_id`) .

src/components/album/AlbumChatBox.js
```diff
@@ -47,28 +47,44 @@ const AlbumChatBox = ({ spotifyUrl }) => {
...
+          const messageWithKey = generateMessageWithKey(data);
           setMessages((prevMessages) => [...prevMessages, messageWithKey]);
-
           scrollToBottom();
         });
       });
       setSocket(newSocket);
     }
+
+    fetchRecentMessages();
```
akosasante: `useEffect` runs everytime any of its dependencies change: https://react.dev/reference/react/useEffect#specifying-reactive-dependencies

For this `useEffect` we have told React that it depends on: `albumId`, `socket`, and `loggedInUserId`.

When the component first loads, it will get the `albumId` by doing regex on the `spotifyUrl` prop. It will also get the logged in user from the `useAuth` hook. But `socket` will be undefined at this point.

This`useEffect` will run for the first time when the component is mounted. Then, because `socket` was undefined we create it and set up all our listeners and whatnot, and then set the socket on line `84`. This means the `socket` variable has changed from `undefined` -> a populated value. This triggers this `useEffect` to run again.

This time, the code inside `albumId && !socket` won't run again since `socket` now has a value; but it **will** run `fetchRecentMessages` again.

We can see this in our dev tools that the messages are being fetched 2 times from the chat endpoint:
Screenshot 2023-11-06 at 2 43 15 PM

---

It's not a huge deal right now but as our site grows, that means every user who enters the chat room, their browser is going to make 2 calls to the backend server, when 1 would have been enough. This wastes our server resources and means we are using more of the user's bandwidth than we need to.

Since we only want to fetch messages, when the user first loads the components, and after that we can rely on the live messages from the socket, we can move the `fetchMessages` code into a separate `useEffect` that has an empty dependencies array so that it only runs when the component is first loaded. 

Alternatively, we can add an `if` statement here so that `fetchMessages` only runs when the socket is not yet setup. The downside with that is we're starting to have a large `useEffect` that is doing many different things. So it's a good sign to start breaking it up in to smaller `useEffect`s

src/components/album/AlbumChatBox.js
akosasante, line 1: line 63, where we set `messageKey` should be no longer needed 👍

src/components/album/AlbumChatBox.js
```diff
@@ -78,6 +78,23 @@ const AlbumChatBox = ({ spotifyUrl }) => {
     };
   }, [albumId, socket, loggedInUserId]);
 
+  useEffect(() => {
+    const fetchRecentMessages = async () => {
+      if (albumId && !socket) {
+        const response = await fetch(
+          `${process.env.REACT_APP_API_BASE_PATH}/chat/${albumId}`
+        );
+        const data = await response.json();
```
akosasante: We should prefer to use `axiosInstance` from the `apis` folder. It's fine overall to use fetch but:
1. we want to try to stick to 1 consistent approach throughout the app. Elsewhere in the app we're using axios, so better to continue following that practice
2. Our axios instance is set up with the BASE_PATH already, so it lets us not need to add it to the url everywhere
3. Our axios instance is set up to log errors and send a cookie if the user is signed-in so that if the backend route requires a logged in user, we don't have to do any extra work on the frontend code.

src/components/album/AlbumChatBox.js
```diff
@@ -78,6 +78,23 @@ const AlbumChatBox = ({ spotifyUrl }) => {
...
+        const messagesWithKeys = data.messages.map((message) => ({
+          ...message,
+          user: { name: message.user },
```
akosasante: Looks good; I'm guessing we may need to change this once you've updated the backend to send the full user object for this endpoint?

akosasante (commented):
looking good; couple small remaining comments

src/components/album/AlbumChatBox.js
mvstoyan: Removed

src/components/album/AlbumChatBox.js
mvstoyan: Do I need to import the User model in handleAlbumChat.js to fetch all the user data on the backend? Or should I add the name directly to the ChatMessage model? Although it doesn't make sense to include it in ChatMessage.js when we already have the data in the User model. Now i don't have a name

mvstoyan: Sorry for my recent "sleepy commits"😅

akosasante: Since `ChatMessage` schema tells mongoose that `user` field is a "ref" (reference) to the `User` model, we can use the `populate` function. It will be the very same as how you've done it in the chat socket to populate the user name here: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/7d57b27f679d6480f43d147f86dcbd305e8b9347/src/live/handleAlbumChat.js#L12

Only you'd be doing it in the `chatRoutes` file where you fetch the chat messages.

src/components/album/AlbumChatBox.js
```diff
@@ -78,6 +77,21 @@ const AlbumChatBox = ({ spotifyUrl }) => {
...
   }, [albumId, socket, loggedInUserId]);
 
+  useEffect(() => {
+    const fetchRecentMessages = async () => {
+      if (albumId && !socket) {
+        const { data } = await axiosInstance.get(`/chat/${albumId}`);
+        console.log('Data from server:', data);
+        const messagesWithKeys = data.messages.map((message) => ({
+          ...message,
+          user: { name: message.user.name, id: message.user._id },
```
akosasante: I think now that the messages are coming in this shape already:

```json
{
    "_id": "6550fd4943fb620d4428ad30",
    "albumId": "6r1lh7fHMB499vGKtIyJLy",
    "message": "good and you?",
    "user": {
        "_id": "64f7d0f615a69acfaa074420",
        "name": "Akos999",
        "id": "64f7d0f615a69acfaa074420"
    },
    "timestamp": "2023-11-12T16:28:57.335Z",
    "__v": 0
}
```

I don't think we need line 86-87 where you're manually setting those values? 🤔  I think we can just pass in messages as-is

akosasante (approved):
Small cleanup suggestion but otherwise looks good to me 👍 🚢

src/components/album/AlbumChatBox.js
mvstoyan: > I think now that the messages are coming in this shape already:
> 
> ```json
> {
>     "_id": "6550fd4943fb620d4428ad30",
>     "albumId": "6r1lh7fHMB499vGKtIyJLy",
>     "message": "good and you?",
>     "user": {
>         "_id": "64f7d0f615a69acfaa074420",
>         "name": "Akos999",
>         "id": "64f7d0f615a69acfaa074420"
>     },
>     "timestamp": "2023-11-12T16:28:57.335Z",
>     "__v": 0
> }
> ```
> 
> I don't think we need line 86-87 where you're manually setting those values? 🤔 I think we can just pass in messages as-is


removed this code

mvstoyan:
@akosasante I encountered an issue with CI tests and had to edit package.json multiple times. Eventually, I reinstalled Prettier using npm install --save-dev prettier. It's possible that I accidentally deleted it in the main branch. I reverted all changes, but I think I should reinstall Prettier in main since it gave me an error. However, I'm worried I might take the wrong steps. Can you please check my branch again?

akosasante:
> @akosasante I encountered an issue with CI tests and had to edit package.json multiple times. Eventually, I reinstalled Prettier using npm install --save-dev prettier. It's possible that I accidentally deleted it in the main branch. I reverted all changes, but I think I should reinstall Prettier in main since it gave me an error. However, I'm worried I might take the wrong steps. Can you please check my branch again?

I don't see it missing in `main` branch: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/package.json#L70

https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/package-lock.json#L25152-L25155

Looking through the failed CI runs I'm not 100% sure why some of those 'install dependencies' steps were failing 🤔 : https://github.com/Code-the-Dream-School/dd-prac-team4-front/actions

But your current change in this PR just sets the `prettier` dependency to any version 3.0.0 or up; rather than strictly `3.0.0`. I think that's a benign enough change. 

Feel free to merge; sorry I can't really provide much clarification here 😓

mvstoyan:
> > @akosasante I encountered an issue with CI tests and had to edit package.json multiple times. Eventually, I reinstalled Prettier using npm install --save-dev prettier. It's possible that I accidentally deleted it in the main branch. I reverted all changes, but I think I should reinstall Prettier in main since it gave me an error. However, I'm worried I might take the wrong steps. Can you please check my branch again?
> 
> I don't see it missing in `main` branch: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/package.json#L70
> 
> https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/package-lock.json#L25152-L25155
> 
> Looking through the failed CI runs I'm not 100% sure why some of those 'install dependencies' steps were failing 🤔 : https://github.com/Code-the-Dream-School/dd-prac-team4-front/actions
> 
> But your current change in this PR just sets the `prettier` dependency to any version 3.0.0 or up; rather than strictly `3.0.0`. I think that's a benign enough change.
> 
> Feel free to merge; sorry I can't really provide much clarification here 😓

Thank you so much for checking. I was worried that I might have caused issues in the main branch. I'll proceed with the merge then.

PR #62, opened 2023-11-05
PS207-Protect-routes

mvstoyan:


src/App.js
```diff
@@ -16,6 +16,13 @@ function App() {
   const { status } = useAuth();
   const isLoggedIn = status === AuthStatus.LoggedIn;
 
+  const ProtectedRoute = ({ children }) => {
+    if (!isLoggedIn) {
+      return <Navigate to="/signIn" />;
+    }
+    return children;
+  };
+
```
akosasante: Sorry this had slipped my mind when writing the ticket; but I belatedly realized we could have also done this with the `RequireAuth` component provided by `react-auth-context` but this works well as well; feel free to keep this or experiment with `RequireAuth` (though using RequireAuth may require some further changes to the auth settings in `index.js` so 🤷 )

https://github.com/akosasante/react-auth-context#:~:text=Alternatively%2C%20use%20the%20RequireAuth%20component%20in%20your%20router%20to%20wrap%20your%20component%20that%20should%20only%20be%20viewed%20if%20logged%2Din%3A

akosasante (approved):
LGTM!

src/App.js
mvstoyan: I'm trying to do this option, but I think I'm going wrong

src/index.js
```diff
@@ -37,11 +37,11 @@ root.render(
     {/* Use CSS baseline to set common spacing/sizing across all of our styling */}
     <CssBaseline />
     <BrowserRouter>
-      <AuthProvider {...authSettings}>
+      <RequireAuth {...authSettings}>
```
akosasante: This one should still be `AuthProvider`

`AuthProvider` is the "Context Provider" component that allows all the children of this element to have access to the useAuth hook. More about React Contexts here: https://react.dev/reference/react/createContext#provider

```suggestion
      <AuthProvider {...authSettings}>
```

src/index.js
akosasante, line 1: If using `RequireAuth`, then in this file we should make sure to update `loginPath` on line 22 to be the path on the frontend, where we want the user to be redirected to if the `RequireAuth` element finds that they are not logged in (probably `/signIn`).

src/App.js
```diff
@@ -18,30 +19,147 @@ function App() {
...
-        <Route
-          path="/chat"
-          element={
-            <AlbumChat spotifyUrl="https://api.spotify.com/v1/albums/6r1lh7fHMB499vGKtIyJLy" />
-          }
-        />
-        {/* Below route is the catch-all route. It **MUST** be the last route because react-router checks the routes from top-to-bottom */}
-        <Route path="*" element={<PageNotFound />} />
-      </Routes>
+      <AuthProvider>
```
akosasante: Let's keep all the provider elements (like BrowserRouter which provides us with access to react-router hooks, and AuthProvider which provides us with access to auth hooks) in the `index.js` file; just for clearer separation of responsibiliites.

mvstoyan: Hope, I did right😅

src/App.js
```diff
@@ -45,3 +76,88 @@ function App() {
...
+//           }
+//         />
+//         {/* Below route is the catch-all route. It **MUST** be the last route because react-router checks the routes from top-to-bottom */}
+//         <Route path="*" element={<PageNotFound />} />
+//       </Routes>
+//     </>
+//   );
+// }
+
+// export default App;
```
akosasante: 🗑️

akosasante (approved):
Looks great 🚢 !

PR #63, opened 2023-11-08
Clicking forgot password navigate page asks backend send forgot password email PS-161

DYA13:


DYA13:
Akos , I have created ForgotPassword.js file. Could you please check? 
Could you please advice how to implement it further in SignIn.js ?

src/App.js
```diff
@@ -31,6 +32,7 @@ function App() {
         <Route path="/wishlist" element={<WishListView />} />
         <Route path="/checkout/completed" element={<CheckoutComplete />} />
         <Route path="/profile" element={<UserProfile />} />
+        <Route path="/forgotPassword" element={<ForgotPassword />} />
```
akosasante: This path `forgotPassword` doesn't match the url that clicking the "Forgot Password" link in `signIn.js` points to (`forgot-password`).

So right now clicking the link sends me to a 404 page not found page:

src/components/ForgotPassword.js
```diff
@@ -0,0 +1,39 @@
...
+import axiosInstance from '../apis/axiosClient';
+
+function ForgotPassword() {
+  const [email, setEmail] = useState('');
+  const [message, setMessage] = useState('');
+  const [error, setError] = useState('');
+
+  const handleSubmit = async () => {
+    try {
+      await axiosInstance.post('/api/v1/auth/forgot_password', { email });
```
akosasante: Our `axiosInstance` already adds in the `/api/v1/` part so that we don't need to repeat it everywhere 😄 

It does that here by setting the `baseURL` to the value of the environment variable `REACT_APP_API_BASE_PATH` https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/8bb8e768ae08fe4a039a860fff7c3f7e5e570c7f/src/apis/axiosClient.js#L5

That environment variable, should be set for everyone locally to `localhost:8000/api/v1` here: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/8bb8e768ae08fe4a039a860fff7c3f7e5e570c7f/.env

So right now, this code here is causing us to make a call to `http://localhost:8000/api/v1/api/v1/auth/forgot_password`, which of course is a url that does not exist in the backend

Screenshot 2023-11-08 at 9 41 46 AM

src/App.js
DYA13: I have updated the url in App.js 👍 commit eef7763e43d24f59afba91fb406a49e3e1c1208d

src/components/ForgotPassword.js
DYA13: I have updated the path 🙈 commit 22471db67c0948ca4a5cd2f7cbc6549a56b571b2

DYA13: Now I see the page to update the password but it throws an error , should be like this at the moment? 

image

akosasante: What does it show in the network tab for the request made? Or dev console for any error logs? Do you have the backend server running and are there any logs in the terminal for the backend that may help point to how to debug this?

It works on my end.

akosasante (approved):
Code looks good to me; please let me know if you are having issues running it locally and what debugging steps you've tried / what error or log messages you're seeing/

Please merge `main` into this branch first to pull in fixes for the lint/format errors that originated from a separate PR.  Then ensure that there are no further lint or format errors from your own code. Then you should be good to go 🚢

src/components/ForgotPassword.js
DYA13: It is rather weird , when I register , it gives me an error but let me see the page 

image

The backend terminal 

image

DYA13: When I try sign in ,it is like this 

image

PR #64, opened 2023-11-09
Write review

NatachaKey:


src/components/review/WriteReview.js
```diff
@@ -0,0 +1,102 @@
+import React, { useState } from 'react';
+import axiosInstance from '../../apis/axiosClient';
+import { useAuth } from '@akosasante/react-auth-context';
+const WriteReview = ({ albumId, refreshReviews }) => {
+  const [formData, setFormData] = useState({
+    user: null,
+    album: null,
+    title: '',
+    rating: 0,
```
akosasante: Does it make sense to default to 0 if it's not a valid number to enter?

If we want it to start off empty, we can start it as `''` (yes we want to use empty string, even though the input element's type is set as number)

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +29,51 @@ const AlbumPreview = ({
...
+          {...props}
+        />
+
+        {children && numOfReviews === 0 ? (
+          <p>No reviews available for this album</p>
+        ) : (
+          <div style={{ width: '100%', display: 'flex' }}>{children}</div>
+        )}
+      </CardContent>
+      <WriteReview />
```
akosasante: By having this set statically in `<AlbumPreview>` component, that means that anywhere this component is used, we will see the `<WriteReview>` component as well. This prevents us from being able to reuse this component on other pages like `http://localhost:3000/chat` where we also want to have a nice Album Preview image + clip player. 

Screenshot 2023-11-10 at 11 43 50 AM

---

Right now, all the content inside of this preview card is nested in a container, `<CardContent>` which is basically a `div` with a bunch of classes. The current element is styled on line 56 to use `Flex` display style. By default, `Flex` displays it's children horizontally. But we can add `flexDirection: 'column'` to that style declaration to have it display its children vertically. (CSS Flexbox Guide: https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

We can (should) even make it a variable that we get from the `props`, that way the caller can be in charge of deciding how they want the album preview to be stacked with its children. 

---

Then, we should pass in all of our review-related code in the `children` component so that this component is not statically always rendering review code. It is able to be "composable" based on what the caller passes in.  In the end, this component should not need to know anything about chat or anything about reviews. It is only concerned with rendering the spotify preview iframe and rendering any children it was passed.

src/components/album/Album.js
akosasante, line 1: On line 72 we can adjust the size of the modal (Dialog component in material-ui) by adding some props/attributes: https://mui.com/material-ui/react-dialog/#optional-sizes

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,47 @@
...
+          {reviews.map((review) => (
+            <li key={review._id}>
+              <h3>{review.title}</h3>
+              <p>Rating: {review.rating}</p>
+              <p>{review.comment}</p>
+            </li>
+          ))}
+        </ul>
+      )}
+    </div>
```
akosasante: I would rather here do:
`{ loading && <p>Loading reviews...</p>}`

and then as a separate check, we can either list the reviews or show the empty state:

`{ reviews?.length ? ( <ul> ....etc </ul> ) : "No reviews yet" }`

One nice thing is that means all our code for loading, populated, and empty state lives in the same component instead of the empty state being all the way in `AlbumPreview.js` as it is right now. 

---

Another approach that some prefer is to pull the conditionals into their own render code instead of having lots of nested ternaries inside the one return statement. This could look like:

```js
 // all our state and use effect stuff is still here, I'm just leaving it out for the sake of space

if (loading) return <p>Loading reviews...</p>
if (!reviews?.length) return <p>No reviews yet</p>

return (
  // reviews.map etc
)
```

akosasante, line 74: Here is where we should render `WriteReview` component.

This is a better place for it because then we can also access any logic we need about reviews like:
- check if user is even logged in; if not, they should not see WriteReview component (but can still read reviews)
- check if any of the reviews we got were by this user, if so, they should not see the WriteReview component (because you are not allowed to write two reviews for the same comment) but instead can edit their review (can be implemented in a separate PR).

src/components/album/Album.js
```diff
@@ -30,14 +32,28 @@ const AlbumTitle = styled('h3')({
...
+    // Set reviewsUpdated to true to trigger a re-render of AlbumReviews
+    setReviewsUpdated(true);
+  };
+
+  useEffect(() => {
+    // If reviews are updated, reset the state after rendering AlbumReviews
+    if (reviewsUpdated) {
+      setReviewsUpdated(false);
+    }
+  }, [reviewsUpdated]);
```
akosasante: I'm not sure that I understand the code flow you're aiming to accomplish here

src/components/album/AlbumPreview.js
NatachaKey: @akosasante Akos, here 
`We can (should) even make it a variable that we get from the props, that way the caller can be in charge of deciding how they want the album preview to be stacked with its children.`
 do you mean we should add this properties using  useState hook like on the screen?
image
?

akosasante: No, not a `useState` hook variable. I mean that currently `AlbumPreview` takes these props: `spotifyUrl`, `style`, `wide`, `width`, `height`, `frameBorder`, `allow`, and `children` (children is a prop that _all_ React components have). 

These are currently mainly used to style the `iframe` element speciifcally. But we could add another prop that this component would accept called like `childrenStackDirection` or some better name. And then we can use that in the `CardContent` element's `style` attribute to set the `flexDirection` CSS value.

akosasante: That way, on the chat page where we want the iframe to stack horizontally with the chat component children, we can pass in `<AlbumPreview  childrenStackDirection='row'...etc />` and on the album modal where we want it to stack vertically with the review component, we can pass in `<AlbumPreview childrenStackDirection='column' ...etc />`

NatachaKey: @akosasante  I think I got it, just to make sure-normally we werepassing props from parent to a child-and here in this case we are destructuring these properties from props but we don't pass it  from anywhere, we only pass the first three from Album.js ,
```
 albumId,
  spotifyUrl,
  numOfReviews,
```
 so if the rest of the props are being assigned directly in props, they are not necessarily have to come from another component, right?
image
image

src/components/album/Album.js
NatachaKey: @akosasante tried to do it, not sure if works ok

src/components/review/ReviewsList.js
NatachaKey: @akosasante thank you, just deleted  this state message 'No reviews' from  AlbumPreview.js is that correct?
image

src/components/album/Album.js
NatachaKey: @akosasante oh I was trying to add code for this task from PS-78 https://codethedreamteam4-2023.atlassian.net/browse/PS-78  **After submitting, should call the refreshReviews function from parent. In parent component, this should trigger re-fetching the reviews for the page **

src/components/album/Album.js
```diff
@@ -30,15 +31,41 @@ const AlbumTitle = styled('h3')({
...
+    setReviewsUpdated(false);
   };
 
+  const handleMaxWidthChange = (event) => {
+    setMaxWidth(event.target.value);
+  };
+
+  const handleFullWidthChange = (event) => {
+    setFullWidth(event.target.checked);
+  };
```
akosasante: Will something be calling these functions? Do we expect to have the user do something to change the max width / full width values? It seems to me like these can just be static variables with no `useState` and no event handlers. But it's possible I'm misunderstanding what you're aiming to do here

src/components/album/AlbumPreview.js
akosasante: The way this component is written, it requires the caller (parent component) to pass in: `albumId`, `spotifyUrl`, and `numOfReviews` otherwise those variables will be `undefined` which may or may not cause errors depending on how those variables are used.

For `style`, `wide`, `width`, `height`, `frameBorder`, and `allow`, the function that makes this component has default values set in the arguments for these props. So even if the parent component does not pass those in, they will be initialized with those default values.

This function component also uses the `rest` parameter (aka destructuring with `...`) to say "okay any other props that are passed in, just save it into one object variable called `props`", and it passes those into the `iframe`. So we should keep in mind that anything a parent component passes in, that is not explicitly named in those arguments, will end up getting passed to the `iframe` which may or may not be what we want.

My suggestion is to add an argument explicitly to that list for determining the flex layout, and then if it has a default value, that means anywhere where want to follow that default value, we do not need to explicitly name that prop in the parent component, but anywhere where we ant to override it we will need to pass that prop in from the parent component.

src/components/review/ReviewsList.js
akosasante: Yes; currently both places that use `<AlbumPreview>` use it and pass in `children` (nested components) but if we want this component to be useable without that, then maybe we should keep the `{children && <div style={{...}}>{children}</div>` part at least; so that the div for the children is only rendered if any children were actually passed in.

src/components/album/Album.js
akosasante: The parent component can pass in a `refreshReviews` function, for this component to call, but I don't see the need to have a separate state `setReviewsUpdated` that we then follow on every render with `useEffect` to accomplish this.

src/components/album/Album.js
```diff
@@ -30,15 +31,41 @@ const AlbumTitle = styled('h3')({
...
+
+  // Function to refresh reviews
+  const refreshReviews = () => {
+    // Set reviewsUpdated to true to trigger a re-render of AlbumReviews
```
akosasante: I'm not understanding how this re-renders the album reviews?

src/components/album/Album.js
```diff
@@ -53,23 +80,40 @@ function Album({ album, wishListId }) {
...
           </Button>
         </DialogActions>
         <DialogContent>
-          <DialogContentText>
+          <DialogContentText  style={{
+          
+          display: 'flex',
+          flexDirection: 'column',
+          justifyContent: 'center',
+          alignItems: 'center',
```
akosasante: Screenshot 2023-11-13 at 10 21 37 AM

I'm not super sure about this style of having _everything_ centered in the modal; it seems to leave a lot of white space 🤔

src/components/album/Album.js
```diff
@@ -53,23 +80,40 @@ function Album({ album, wishListId }) {
...
               </div>
               <div>Average Rating: {album.averageRating}</div>
             </div>
           </DialogContentText>
-          <AlbumPreview spotifyUrl={album.spotifyUrl} />
         </DialogContent>
+        <AlbumPreview
+          albumId={album._id}
+          spotifyUrl={album.spotifyUrl}
+          numOfReviews={album.numOfReviews}
```
akosasante: Since we're trying to avoid giving `AlbumPreview` information it doesn't need; let's not pass in this prop here. Instead, it probably makes more sense to give this information to `AlbumReviews` component below and have it render the empty state message if needed.

src/components/album/AlbumPreview.js
```diff
@@ -1,9 +1,13 @@
 import React, { useEffect, useState } from 'react';
 import { CardContent } from '@mui/material';
-
+import axiosInstance from '../../apis/axiosClient';
+import WriteReview from '../review/WriteReview';
 const AlbumPreview = ({
+  albumId,
   spotifyUrl,
+  numOfReviews,
```
akosasante: Per previous comment, let's move this to AlbumReview since it's no longer being used here

src/components/album/AlbumPreview.js
```diff
@@ -13,6 +17,7 @@ const AlbumPreview = ({
   ...props
 }) => {
   const [spotifyEmbedUrl, setSpotifyEmbedUrl] = useState('');
+  const [mongoAlbumId, setMongoAlbumId] = useState(''); // State to store MongoDB album ID
```
akosasante: Do we need this still?

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +30,65 @@ const AlbumPreview = ({
...
+    axiosInstance
+      .get(`/albums/${albumId}`)
+      .then((response) => {
+        console.log(response.data.album._id);
+
+        setMongoAlbumId(response.data.album._id);
+      })
+      .catch((error) => {
+        console.error('Error fetching MongoDB album ID:', error);
+      });
```
akosasante: I believe we can remove this code now?

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +30,65 @@ const AlbumPreview = ({
...
-    </CardContent>
+    <>
+      <CardContent
+        style={{
+          marginTop: '1rem',
+          width: '100%',
+          display: 'flex',
+          flexDirection: contentLayout,
+          justifyContent: 'center',
+          alignItems: 'center',
```
akosasante: Do we need everything centered horizontally and vertically; or just the children div?

src/components/album/AlbumPreview.js
akosasante: Or maybe we don't need it there anyway if we can check the review api response length instead 😄

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,51 @@
...
+        const response = await axiosInstance.get(`/reviews/album/${albumId}`);
+        console.log(response.data.allProductReviews);
+
+        const { allProductReviews } = response.data;
+        setReviews(allProductReviews);
+        setLoading(false);
+      } catch (error) {
+        console.error('Error fetching album reviews:', error);
+        setLoading(false);
+      }
```
akosasante: Maybe we want to display some sort of error message to the user (in the div or maybe as an Alert/Snackbar component) if the fetch fails?

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,51 @@
...
+              <h3>{review.title}</h3>
+              <p>Rating: {review.rating}</p>
+              <p>{review.comment}</p>
+            </li>
+          ))}
+        </ul>
+      ) : (
+        'No reviews yet'
+      )}
+      <WriteReview/>
```
akosasante: This component is expecting the `albumId` and a `refreshReviews` function in its props but here we're not passing anything in. That makes the Submit Review button fail because it will try to make a POST call to `http://localhost:8000/api/v1/reviews/album/undefined`

Screenshot 2023-11-13 at 10 28 01 AM

akosasante: This component can define a refreshReviews button which when called, reruns the `fetchAlbumReviews` function and resets the `reviews` variable which will cause this component to rerender the list of reviews and any children it has (WriteReview).

src/components/album/Album.js
NatachaKey: @akosasante sure! you are right, it's better to have it as a variable- I was just thinking we might want this modal to have diferent sizes depending on where we call it

src/components/album/Album.js
NatachaKey: @akosasante oh that task was not fiished.. I wasn't sure how to do it without useState and useEffect hooks

src/components/album/AlbumPreview.js
```diff
@@ -1,9 +1,12 @@
 import React, { useEffect, useState } from 'react';
 import { CardContent } from '@mui/material';
-
+import axiosInstance from '../../apis/axiosClient';
+import WriteReview from '../review/WriteReview';
 const AlbumPreview = ({
+  albumId,
```
akosasante: I don't think this is needed any longer with latest changes

src/components/album/AlbumPreview.js
```diff
@@ -1,9 +1,12 @@
 import React, { useEffect, useState } from 'react';
 import { CardContent } from '@mui/material';
-
+import axiosInstance from '../../apis/axiosClient';
+import WriteReview from '../review/WriteReview';
```
akosasante: no longer used with latest changes

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +28,60 @@ const AlbumPreview = ({
...
-    const albumId = match[1];
+    const spotifyAlbumId = match[1];
+
     //  Spotify album URL for embedding
-    const spotifyAlbumUrl = `https://open.spotify.com/embed/album/${albumId}`;
+    const spotifyAlbumUrl = `https://open.spotify.com/embed/album/${spotifyAlbumId}`;
     setSpotifyEmbedUrl(spotifyAlbumUrl);
   }, [spotifyUrl]);
-
+  //console.log("Children:", children.type);
```
akosasante: can be removed once we're ready to merge

src/components/album/Album.js
```diff
@@ -53,23 +57,40 @@ function Album({ album, wishListId }) {
...
             <div>
               <div>
                 Release Date: {new Date(album.releaseDate).toLocaleDateString()}
               </div>
               <div>Average Rating: {album.averageRating}</div>
             </div>
           </DialogContentText>
-          <AlbumPreview spotifyUrl={album.spotifyUrl} />
         </DialogContent>
+        <AlbumPreview>
```
akosasante: `<AlbumPreview>` component expects a `spotifyUrl`  property to be passed in. Otherwise, it's `useEffect` will fail on line 24 of `AlbumPreview.js` when it tries to do a regex on that property, if that property is undefined.

akosasante: This is where adding propTypes to components can come in handy for helping the developer (and the linter) see what props on a component are required vs optional. But we can circle back to that later :slightly\_smiling\_face:

src/components/album/AlbumPreview.js
```diff
@@ -1,9 +1,12 @@
 import React, { useEffect, useState } from 'react';
 import { CardContent } from '@mui/material';
-
+import axiosInstance from '../../apis/axiosClient';
+import WriteReview from '../review/WriteReview';
 const AlbumPreview = ({
+  albumId,
   spotifyUrl,
   style = {},
+  contentLayout = 'column',
```
akosasante, line 7: Since we've made this by default `"column"`. We should make sure we don't break how the `localhost:3000/chat` page was coded, where it wanted the album preview to be side by side (content layout: "row") with the chat box (or check with @mvstoyan  if we're fine with the design of that page changing to a column layout). 

If we want to make sure the chat page continues to be rendered in horizontal fashion, then we should update `AlbumChat.js` to pass in `contentLayout="row"` to the `<AlbumPreview>` component that it renders.

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +28,60 @@ const AlbumPreview = ({
...
+          }}
+        >
+          {children}
+        </div>
+
+        {/* {children && numOfReviews === 0 ? (
+          <p>No reviews available for this album</p>
+        ) : (
+          <div style={{ width: '100%', display: 'flex' }}>{children}</div>
+        )} */}
```
akosasante: This can be removed once we're ready to merge :+1:

src/components/album/AlbumPreview.js
```diff
@@ -25,28 +28,60 @@ const AlbumPreview = ({
...
+          allow={allow}
+          {...props}
+        />
+
+        <div
+          style={{
+            width: '100%',
+            display: 'flex',
+            flexDirection: contentLayout,
+            justifyContent: 'center',
```
akosasante, line 70: Since this div will only have one child element, (whatever is rendered by `children`) keep in mind that the flex properties here only apply to those children. You can play around with the values in the dev console; but the main thing I want to point out that just because we want the items in `<CardContent>` to be either "column" or "row" depending on what the parent component passes in; doesn't necessarily mean we wnat that same value for within each of the children (iframe, this div). So maybe it doesn't make sense to pass in that same prop to all these children, vs hardcoding some values there.

src/components/album/Album.js
```diff
@@ -30,6 +31,9 @@ const AlbumTitle = styled('h3')({
 
 function Album({ album, wishListId }) {
   const [openDialog, setOpenDialog] = useState(false);
+  const reviewsRef = React.useRef(); // Created a ref to hold the function
```
akosasante: I'm not understanding the use-case for this line?

src/components/album/Album.js
```diff
@@ -30,6 +31,9 @@ const AlbumTitle = styled('h3')({
...
+  const fullWidth = true;
+  const maxWidth = 'sm';
```
akosasante: Will these be used in multiple places, or can we just inline them on the lines where they are being used for the `<Dialog>` component?

src/components/album/Album.js
```diff
@@ -53,23 +57,40 @@ function Album({ album, wishListId }) {
...
               <div>Average Rating: {album.averageRating}</div>
             </div>
           </DialogContentText>
-          <AlbumPreview spotifyUrl={album.spotifyUrl} />
         </DialogContent>
+        <AlbumPreview>
+          <AlbumReviews
+            albumId={album._id}
+            spotifyUrl={album.spotifyUrl}
+            numOfReviews={album.numOfReviews}
```
akosasante: `AlbumReviews.js` only shows the component accepting one prop: `albumId`. Are these other two (`spotifyUrl` and `numOfReviews` expected to be used by the `AlbumReviews` component for anything?

src/components/album/Album.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
```
akosasante: I think `useEffect` is no longer being used for anything in this file.

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,61 @@
...
+      setLoading(false);
+    }
+  };
+
+  // Function to refresh reviews
+  const refreshReviews = () => {
+    // Simply call the fetchAlbumReviews function to refresh
+    fetchAlbumReviews();
+  };
+
```
akosasante:
```suggestion
  const refreshReviews = () => {
    setLoading(true) // show the loading text while we go to fetch/refresh the reviews
    // Simply call the fetchAlbumReviews function to refresh
    fetchAlbumReviews();
  };
```

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div>
+      <h2>Album Reviews</h2>
+      {error && <Alert severity="error">{error}</Alert>}
+
+      {loading && <p>Loading reviews...</p>}
+
+      {reviews?.length ? (
+        <ul>
+          {reviews.map((review) => (
```
akosasante, line 55: It might be nice to show when the review was written and the user name of the reviewer. But these could maybe be extra things to implement in a separate PR, just an idea.

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,61 @@
...
+              <h3>{review.title}</h3>
+              <p>Rating: {review.rating}</p>
+              <p>{review.comment}</p>
+            </li>
+          ))}
+        </ul>
+      ) : (
+        'No reviews yet'
+      )}
+      <WriteReview albumId={albumId} refreshReviews={refreshReviews} />
```
akosasante: This could be left to a future PR, but keep in mind that we only want to show the `WriteReview` component if the user has not already reviewed this album, and only if the user is actually logged in at all.

src/components/review/WriteReview.js
```diff
@@ -0,0 +1,102 @@
...
+        <div>{error && <p style={{ color: 'red' }}>{error}</p>}</div>
+      </form>
+    </div>
+  );
+};
+
+export default WriteReview;
+
+//Akos: I had to hardcode the userId- not sure where to retrieve it from?
+//need to check later id this option 'write a review' is shown only to users who didn't leave a review for this/that album yet
```
akosasante: I think you have this now :+1:

src/components/review/WriteReview.js
```diff
@@ -0,0 +1,102 @@
...
+  const [formData, setFormData] = useState({
+    user: null,
+    album: null,
+    title: '',
+    rating: '',
+    comment: '',
+  });
+  const [submitting, setSubmitting] = useState(false);
+  const [error, setError] = useState(null);
+  const { user } = useAuth();
```
akosasante: console log this value; I think the actual user object you want is a little nested in there. This is something we can fix/simplify in some future work; but for now, you need to do `user.user.<whatever field you want>` to access things properly

src/components/album/AlbumPreview.js
akosasante: https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/119d9dce-90ec-4693-9cb4-b897d37fefa3

src/components/album/Album.js
NatachaKey: @akosasante go you refer to the use of this prop-types package?  //couldn't implement it without getting errors
image

src/components/album/AlbumPreview.js
NatachaKey: @akosasante sure, my bad, will leave it in a horizontal way

src/components/album/Album.js
NatachaKey: just forgot to delete it, was trying to do smth

src/components/album/Album.js
NatachaKey: @akosasante they don't.. will delete them... to avoid confusion I realize I should have named AlbumReviews component as ReviewsList.js as the file's name,..

src/components/review/ReviewsList.js
NatachaKey: @akosasante great idea . will implement it in another PR

src/components/review/ReviewsList.js
NatachaKey: @akosasante just added code for this please take a look

src/components/album/AlbumChat.js
```diff
@@ -2,7 +2,7 @@ import AlbumPreview from './AlbumPreview';
 import AlbumChatBox from './AlbumChatBox';
 
 const AlbumChat = ({ spotifyUrl }) => (
-  <AlbumPreview spotifyUrl={spotifyUrl} wide>
+  <AlbumPreview spotifyUrl={spotifyUrl} wide contentLayout="row">
```
akosasante, line 5: :pray::+1:

src/components/album/AlbumPreview.js
```diff
@@ -4,6 +4,7 @@ import { CardContent } from '@mui/material';
 const AlbumPreview = ({
   spotifyUrl,
   style = {},
+  contentLayout,
```
akosasante: To clarify my previous comment, we still may want the default value here otherwise, it'll be `undefined` the way you're calling it in `Album.js` with no contentLayout passed in (eg: `<AlbumPreview spotifyUrl={album.spotifyUrl}>`). 

My previous comment was to ensure that you override the default for `AlbumChat` so that we don't accidentally mess with how that page is rendered. 

```suggestion
  contentLayout = "column",
```

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,75 @@
...
+
+      {loading && <p>Loading reviews...</p>}
+
+      {reviews?.length ? (
+        <ul>
+          {reviews.map((review) => (
+            <li key={review._id}>
+              <h3>{review.title}</h3>
+              <p>Rating: {review.rating}</p>
+              <p>{review.comment}</p>
```
akosasante: You may want to consider adding an `overflow-wrap` css style here since the review comment could potentially be very long

src/components/review/ReviewsList.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const fetchAlbumReviews = async () => {
+    try {
+      const response = await axiosInstance.get(`/reviews/album/${albumId}`);
+      console.log(response.data.allProductReviews);
+
+      const { allProductReviews } = response.data;
+      setReviews(allProductReviews);
+      // Check if the user has already reviewed the album
+      const hasReviewed = allProductReviews.some(review => review.user === user.user._id);
```
akosasante: This will cause errors if `user` is null (ie if a non-logged in user wants to take a look at reviews -- the reviews load fine, and the setReviews call happens sucessfully before this line, but then an error is caught here which causes the error message to be shown even though the reviews were loaded completely fine). We can put this in an `if` statement or use optional chaining on `user?.user?._id`

akosasante (commented):
Couple small remaining notes; looking great!

src/components/album/AlbumPreview.js
NatachaKey: @akosasante to be honest, I am lost a bit with the hierarchy- which component is the parent and passes props to it's children. I see it as `AlbumList.js` fetches all the albums and passes them to `AlbumGrid.js` , this one  passes '**each'album** to  `Album.js` component  that displays `AlbumPreview.js` and `AlbumReviews `- the first one is basically the card /design that shows it's content- so if `Album.js `is the parent component for `AlbumPreview.js`, and we don't pass contentLayout , so we should pass it directly along with props = declare it for the first time, right? here on line 7 (img above) so that we could use it when render the page- right?
image

src/components/review/ReviewsList.js
NatachaKey: @akosasante Thank you- with your changes everything works great! 
I am just wondering.. because
I've made sure I'm logged in and it throws an error (with no changes in code)- I just don't get why it sets user to _null_.

src/components/album/Album.js
akosasante: Hmm, maybe pull in the latest from `main` I think we did already implement some prop types in some places: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/945fedf11c6a8694d6d0f7b04dcfcb366f3e2133/src/components/album/Album.js#L80-L83](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/945fedf11c6a8694d6d0f7b04dcfcb366f3e2133/src/components/album/Album.js#L80-L83)

And yes we were using the `prop-types` package (lately React officially recommends using that one). You may need to install that in your branch if it's not available.

src/components/review/ReviewsList.js
akosasante: looks great :+1:

src/components/review/ReviewsList.js
akosasante: Could you show the error it threw before? It could be that when the component first runs the `user` variable hasn't yet been loaded. React will re-render the component automatically when any of its props or variables its getting from hooks or `useState` change.

akosasante (approved):
LGTM. Let me know if I missed any of your questions above.

PR #67, opened 2023-11-13
Create a page where you can update your password, calling the reset password endpoint PS-163

DYA13:


DYA13:
I have started working with this ticket but I need your guidance, could you please help how  to proceed on  what I have done and how to test?

akosasante (commented):
One thing to fix in the backend. Currently the backend email we generate for the user to reset their password points to `http://localhost:8000/resetPassword?token=<token>` because at the time the plan was to maybe create that page as a EJS server-rendered page. But since we're now creating this page in the _frontend_ we should update the `resetLink` that's generated here to point to the _frontend_ (which in local dev the environment variable should be `http://localhost:3000` and in production (in Render.com) the environment variable should be `https://beatbazaar.onrender.com/`): https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/6af16ef2609bbf8a67a4e4d744c1f7d3316ddd6b/src/mailing/sender.js#L83

src/App.js
```diff
@@ -32,6 +34,8 @@ function App() {
         <Route path="/wishlist" element={<WishListView />} />
         <Route path="/checkout/completed" element={<CheckoutComplete />} />
         <Route path="/profile" element={<UserProfile />} />
+        <Route path="/forgot-password" element={<ForgotPassword />} />
+        <Route path="/forgot-password" element={ResetPassword} />
```
akosasante: We can't use the same path for two different routes. How will React know which component to render?

akosasante: I think currently it will try to render the first one only; but that means we have no way to get to the Reset Password page.

src/apis/authApi.js
akosasante, line 1: Thank you for creating this API code in the api folder! We have some work to do to follow this pattern throughout the codebase 🤞

src/apis/authApi.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const resetPassword = async ({ token, password }) => {
+  try {
+    const response = await axios.post('/api/v1/auth/forgot_password', {
+      token,
+      password,
+    });
+    return response.data;
+  } catch (error) {
+    throw error;
```
akosasante: Catching an error to only re-throw it again is a bit redundant. Without the try catch, the code would basically do the same thing anyway. Catch is useful if we're going to do something with the error like log it or use it to change state or something.

src/apis/authApi.js
```diff
@@ -0,0 +1,13 @@
+import axios from './axiosClient';
+
+export const resetPassword = async ({ token, password }) => {
+  try {
+    const response = await axios.post('/api/v1/auth/forgot_password', {
```
akosasante: `forgot_password` is the endpoint for sending the forgot password email; we're already calling this endpoint in `ForgotPassword.js`

For _resetting_ the password we have a separate route in the backend (taking the user's new password and updating it in the database): https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/6af16ef2609bbf8a67a4e4d744c1f7d3316ddd6b/src/routes/authRoutes.js#L16

src/components/ResetPassword.js
```diff
@@ -0,0 +1,54 @@
+// ForgotPassword.js
+import React, { useState } from 'react';
+import { useHistory, useLocation } from 'react-router-dom';
+import { resetPassword } from '../apis/authApi'; // Make sure to import your API call function
+
+const ResetPassword = () => {
+  const history = useHistory();
```
akosasante: `useHistory` was deprecated after React Router v5 (we are using v6 if you check our package.json): https://github.com/remix-run/react-router/blob/main/docs/upgrading/v5.md#use-usenavigate-instead-of-usehistory

akosasante: This error when trying to navigate to this component is saying that `useHistory` is not a valid variable/function exported from the `react-router-dom` library:

Screenshot 2023-11-13 at 10 01 21 AM

src/App.js
akosasante: Screenshot 2023-11-13 at 10 02 31 AM

`ResetPassword` by itself is the function returned from the `ResetPassword.js` file (`const ResetPassword = () => {...}` is a function). 

But `Route` component is not expecting a function in the `element` prop, it's expecting a _component_. To make a React function into a component, we need to add the `< ...>` brackets to render it as a DOM element. Note how all the other elements in the other routes are written.

src/apis/authApi.js
akosasante: Our `axiosClient` already adds `/api/v1` to any routes we call so that we don't have to add that to every http call ourselves
https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/page_update_password/src/apis/axiosClient.js#L3C29-L3C52

(it uses `REACT_APP_API_BASE_PATH` as the base path for any http call made using this base axios client which is defined in our env file as `http://localhost:8000/api/v1` for local dev and similarly for the deployed one: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/page_update_password/.env#L1)

Otherwise, right now the code you have here will try to make a call to `http://localhost:8000/api/v1/api/v1/auth/forgot_password`

src/components/ResetPassword.js
```diff
@@ -0,0 +1,54 @@
...
+    try {
+      if (password !== confirmPassword) {
+        setError('Passwords do not match');
+        return;
+      }
+
+      await resetPassword({ token, password });
+
+      // Redirect to sign in page on success
+      history.push('/signin');
```
akosasante: Check how to do this with the `useNavigate` hook for react-router-dom v6

DYA13:
Hello Akos , regarding the first comment for updating resetLink,  should I make a separate branch and PR to implement these changes on backend?

DYA13:
> update
I have changed in backend resetLink , commit 06cd1cff75164de44cfca6f7148e643863faf8e2

src/App.js
DYA13: Sorry , my mistake , I have changed now commit 78b11dba2e2ffe1b94d837a837d636a6045ab353

src/apis/authApi.js
DYA13: corrected 🙈 ,  commit b1d1ccf5b908483e5defa6e1daef3e2d9a06e505

src/apis/authApi.js
DYA13: Updated, thank you , commit 483d19688982bf1d381716d77c3731e520d4e320

src/components/ResetPassword.js
DYA13: Thank you Akos, commit 4266b8e53034aadd3b49d03dbcbf55488e7ecd81

src/App.js
DYA13: Edited , done 👍

src/apis/authApi.js
DYA13: Corrected 👍, commit 1f6dc4e98f593f7aa8e0ad1fde3e767fbcca940d

src/components/ResetPassword.js
DYA13: Thank you , edited👍

src/App.js
```diff
@@ -67,7 +68,8 @@ function App() {
         <Route path="/checkout/completed" element={<CheckoutComplete />} />
         <Route path="/profile" element={<UserProfile />} />
         <Route path="/forgot-password" element={<ForgotPassword />} />
-        <Route path="/updateUserInfo" element={<PersonalProfileEditForm />} />
```
akosasante: Was it intentional to delete this route?

src/apis/authApi.js
```diff
@@ -0,0 +1,9 @@
+import axios from './axiosClient';
+
+export const resetPassword = async ({ token, password }) => {
+  const response = await axios.post('/auth/resetPassword', {
+    token,
+    password,
```
akosasante: The backend `resetPassword` endpoint expects a "passwordToken" field and a "newPassword" field. Currently this is passing in a "token" field and a "password" field so it won't be successful: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/f0cf82f1f4446eaffe3143c03fc372f28d9b8648/src/controllers/authController.js#L184C1-L184C1

src/components/ResetPassword.js
```diff
@@ -0,0 +1,54 @@
...
+    try {
+      if (password !== confirmPassword) {
+        setError('Passwords do not match');
+        return;
+      }
+
+      await resetPassword({ token, password });
+
+      // Redirect to sign in page on success
+      navigate.push('/signin');
```
akosasante: `navigate` doesn't have a `.push` function:

[https://github.com/remix-run/react-router/blob/main/docs/upgrading/v5.md#use-usenavigate-instead-of-usehistory](https://github.com/remix-run/react-router/blob/main/docs/upgrading/v5.md#use-usenavigate-instead-of-usehistory)

^ The example there shows that instead of `history.push` , now in React-Router v6 we can pass the url path we want to navigate to directly into `navigate` as a function call: `navigate("/signIn")`

akosasante:
> > update  
> > I have changed in backend resetLink , commit 06cd1cff75164de44cfca6f7148e643863faf8e2

Is this pushed to Github / do you have a PR open I can review for the backend change?

src/App.js
DYA13: I have deleted in my branch as I had an error in section 'problems' that it is not defined or not imported for passing github actions.

src/apis/authApi.js
DYA13: Thank you , done ,commit  205162a3c8425bacaf3ca25266c1831b5a882052

src/components/ResetPassword.js
DYA13: You are right , sorry, commit bb3122006976aea093071aa9dc2bbbd7abc4278d

DYA13:
> > > update
> > > I have changed in backend resetLink , commit 06cd1cff75164de44cfca6f7148e643863faf8e2
> 
> Is this pushed to Github / do you have a PR open I can review for the backend change?

Yes Akos , sure , https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/88

src/components/ResetPassword.js
```diff
@@ -0,0 +1,54 @@
...
+  const [error, setError] = useState('');
+
+  const handleResetPassword = async () => {
+    try {
+      if (password !== confirmPassword) {
+        setError('Passwords do not match');
+        return;
+      }
+
+      await resetPassword({ token, password });
```
akosasante: The `resetPassword` function is expecting an object with properties named: `passwordToken` and `newPassword`, but here you're passing an object with the properties `token` and `password`. If you stick a console.log in `resetPassword` you'll see that it's not getting those value because of this mismatch.

You can also see in the dev tools in the Network tab that nothing is being sent to the backend:

akosasante: Remember that when we write an object with just one value, it's a shortcut for saying that the `key` and the `value` have the same name. So what you've written is actually doing: `{ token: token, password: password}`

You can either name the keys the way the function expects them here: `await resetPassword({ passwordToken: token, newPassword: password })` .

Or you can update the function to read the keys that you're passing it. In that case, you still need to make sure that the object you're passing as the request body has the keys that the server is expecting:

```javascript
export const resetPassword = async ({ token, password }) => {
  const response = await axios.post('/auth/resetPassword', {
    passwordToken: token,
    newPassword: password,
  });
  return response.data;
};
```

akosasante (approved):
LGTM 🚀

PR #68, opened 2023-11-15
PS165-list-purchased-albums + PS208 (use params for album chat page)

mvstoyan:


mvstoyan:
in progress

src/components/user/UserProfile.js
```diff
@@ -10,10 +10,21 @@ import {
...
+  const [purchasedAlbums, setPurchasedAlbums] = useState([]);
+
+  useEffect(() => {
+    const fetchPurchasedAlbums = async () => {
+      const response = await axiosInstance.get('/users/showMe/withAlbums');
+      console.log(response.data.user.purchasedAlbums);
+      setPurchasedAlbums(response.data.user.purchasedAlbums);
+    };
+    fetchPurchasedAlbums();
+  }, [setPurchasedAlbums]);
```
akosasante: On my user, I have a lot of duplicate albums, because I have purchased the same album lots of times 😅. I'm guessing that should be allowed in case I want multiple copies of the album for whatever reason as a customer. But when showing them here, maybe we should do a `purchasedAlbums.filter` here first to remove any duplicates (or use a helper function from a library like [lodash](https://stackoverflow.com/questions/31740155/lodash-remove-duplicates-from-array) or `underscore`) just to clean up the UI a bit.

src/components/album/AlbumChat.js
```diff
@@ -1,10 +1,12 @@
 import AlbumPreview from './AlbumPreview';
 import AlbumChatBox from './AlbumChatBox';
+import { useParams } from 'react-router-dom';
 
-const AlbumChat = ({ spotifyUrl }) => (
-  <AlbumPreview spotifyUrl={spotifyUrl} wide>
-    <AlbumChatBox spotifyUrl={spotifyUrl} />
+const AlbumChat = () => {
+  const { albumId } = useParams();
+  <AlbumPreview albumId={albumId} wide>
```
akosasante: Keep in mind that `<AlbumPreview>` component is expecting the spotify URL in order to be able to get the spotify API id that it needs to render that Spotify iframe player for playing tracks from the album. Here you've changed it to pass in the album id, but since you haven't changed the actual `AlbumPreview.js` file, I expect that this might break some things.

akosasante: Same with the change below to `AlbumChatBox`. Maybe you do just want to use the album's Mongo id in the chat component, but we should make sure the props we're passing in match the props the component is expecting.

src/App.js
```diff
@@ -94,10 +94,10 @@ function App() {
           }
         />
         <Route
-          path="/chat"
+          path="/chat/:albumId"
```
akosasante, line 97: :+1:

mvstoyan:
I got confused with tickets PS165/208, sorry. But I'm receiving an 'API invalid' error. I can't figure out how to fix it, can you please guide me?

akosasante:
Sorry to clarify, should I be reviewing the PS-165 -- listing purchased albums; or PS-208 -- generalize the chat album URL work in this PR @mvstoyan. Or should we consider this a combo of both?

akosasante:
Debugging "Invalid API URL":

*   dev console shows the "Invalid API URL" and points us to the `if(!match)` section of the code in AlbumPreview.js:



*   After clicking the link in the dev console we can see the surrounding code and guess that the regex to save into the `match` variable didn't work



*   We see that before that step we log the albumId and the match, so let's check the dev console again:



So the question is why did `match` end up as null?

We see in that the following steps are done:

```
    // regular expression pattern to extract the album ID from the Spotify **API** Url that's saved on the album object
    const regexPattern = /\/albums\/([a-zA-Z0-9]+)/;
    //  extract the album ID
    const match = albumId.match(regexPattern);
```

So we are executing a Regex check on `albumId`, expecting to find the text `/albums/<bunch of letters and numbers>`

But we see in our console log, that `albumId` was _not_ a spotify API Url with that pattern, but now with the new code is just the Mongo ID of the album itself (`album._id` field, that was used in the Link component in the UserProfile component and then grabbed from the params in the `AlbumChat` component ([here)](https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/68/files#diff-dc044c6ce14501135532d751cb2c59b15079b67646f1634abe6dd9c739c899b4R6)

The `AlbumPreview` component needs the **Spotify API id** (not our own mongo database id that our server has created) so that we can give that to the iframe that renders the spotify play widget. Since we're not storing that field in our database, but instead are storing the whole API URL, we need to use Regex to extract the part that we need.

For the `AlbumChat` component I think it probably doesn't matter whether we use the Spotify ID or our own Mongo ID, whichever is easier. Since that is just used to define a io.socket room in our own server and to know which chat messages belong to which room/album/conversation.

mvstoyan:
> Sorry to clarify, should I be reviewing the PS-165 -- listing purchased albums; or PS-208 -- generalize the chat album URL work in this PR @mvstoyan. Or should we consider this a combo of both?

I initially thought of tackling both of them together, but it seems like it was a hasty decision that's only causing confusion. I'll make a separate branch for PS-208 and debugging "Invalid API URL" there.

akosasante:
> I initially thought of tackling both of them together, but it seems like it was a hasty decision that's only causing confusion. I'll make a separate branch for PS-208 and debugging "Invalid API URL" there.

Ah I see. I think it's fine to keep them both here in that case, if that's easier than trying to separate them at this point. Just wanted to make sure I was on the same page as you :+1:

akosasante:
Closing this for now; we can still refer to it but maybe best to start a fresh branch -- let me know what you think and if you'd prefer to reopen this PR?

mvstoyan:
> Closing this for now; we can still refer to it but maybe best to start a fresh branch -- let me know what you think and if you'd prefer to reopen this PR?

@akosasante In this PR, there are two tickets: 165 - A component in the user profile to list purchased albums, and ticket 208 - Make the URL for the album chat page variable instead of hardcoded. For ticket 165, the albums are displayed. The error occurs with ticket 208. It would be good to handle ticket 208 separately to avoid confusion. This might help me fix the error when clicking the "Chat with Fans" button. I just don't understand how to do it correctly. If we close it, then 165 also disappears.

akosasante:
@mvstoyan let's make a new separate PR for PS-165, just so that we have a fresh start with this changes now that `main` branch has changed so much. Follow these steps please:

```bash
# checkout main branch
git checkout main

# pull latest changes from main
git pull

# create a new branch for the PS-165 ticket
git checkout -b PS165-new

# We are going to use git's `cherry-pick` command to pick all the commits from this PR and put them on the new branch
# Read more here: https://www.atlassian.com/git/tutorials/cherry-pick
# The "--no-commit" flag means that it will pull the changes from main but won't auto-commit them
# This gives us a better chance to pick just the changes that are relevant to PS165
# The "~" after the first commit SHA, tells Git that we want to include that first commit in the range of commits that we are picking here
#
# Go to the commits in the original PR and pick the first commit, click the copy button and the last commit and copy it to make the command below
# https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/68/commits
git cherry-pick 404f57e8363c07c43e89cddfecff25fdaeec9518~..b5ea7e7120c2d935732f0ba069d93f7b59d7ccd9 --no-commit

# There may be some conflicts to resolve, specifically in the UserProfile.js file
# You should be okay to just accept all your changes for this (on the left in my Vs Code, see below image)
# After resolving the conflicts, only select the "UserProfile.js" file to stage and commit in VS Code
# Can do in VS Code or in terminal with:
git add . && git commit -m "PS165 -- add purchased albums to user profile page"


# Then tell git to continue cherry-picking by proceeding to the next commit
git cherry-pick --continue

# Now we may see some more merge conflicts but for PS165, we only care about UserProfile.js changes
# So only select the changes from UserProfile, and the package/package-lock files to keep
# See my image below for what I mean

# Continue again with git cherry-pick continue, and keep selecting just the PS165 relevant changes until you're all done


# Now push this to a new branch on Github (remote/origin):
git push --set-upstream origin PS165-new

# In the Github UI open a PR for this
```

 -- copy button is this one:



\-- this is what I mean by only selecting UserProfile.js for the PS165 merges (click the "-" minus sign to remove a file from the "staged changes" section so that it doesn't get accidentally committed):

[https://imgur.com/a/00gKKWX](https://imgur.com/a/00gKKWX)

---

Then we can do the same for the PS208 changes:

```bash
# checkout main branch
git checkout main

# pull latest changes from main
git pull

# create a new branch for the PS-165 ticket
git checkout -b PS208-new

# This time we remove the "~" because we don't need the first commit, since it only has PS165 content.
git cherry-pick 404f57e8363c07c43e89cddfecff25fdaeec9518..b5ea7e7120c2d935732f0ba069d93f7b59d7ccd9 --no-commit

# Now this time as you go through each commit, keep all the file changes _except_ for anything to do with UserProfile / PS165 changes
```

PR #69, opened 2023-11-18
A page in the frontend that shows recommendations based on a user's play history PS-211

DYA13:


src/App.js
```diff
@@ -12,11 +12,12 @@ import UserProfile from './components/user/UserProfile';
 import CheckoutComplete from './components/Purchase/CheckoutComplete';
 import PageNotFound from './components/PageNotFound';
 import AlbumChat from './components/album/AlbumChat';
+import RecommendationPage from './components/RecommendationPage';
 import ForgotPassword from './components/ForgotPassword';
 import { createTheme, CssBaseline, ThemeProvider } from '@mui/material';
 
 function App() {
-  const { status } = useAuth();
+  const { status, user } = useAuth();
```
akosasante: This will work fine, but I would suggest instead fetching the current user within the `RecommendationPage` component. My logic being that the logic around what the component wants to render and how it wants to handle having a logged-in user or not all belongs to that component.

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState, useEffect } from 'react';
+
+const RecommendationPage = ({ userId }) => {
+  const [recommendations, setRecommendations] = useState([]);
+
+  useEffect(() => {
+    fetch(`/recommendations/${userId}`)
```
akosasante: Right now, this is trying to make a call to the front-end itself at a path like: `http://localhost:3000/recommendations/undefined`:



That's because you haven't passed in any information to `fetch` for it to know how to call the backend. You could pass in the whole URL you want to call : `http://localhost:8000/api/v1/recommendations` to the fetch call. But then you also need to make sure that when deployed, it calls [`https://musicstore.onrender.com/api/v1/recommendations`](https://musicstore.onrender.com/api/v1/recommendations) instead.  To help with not having to remember to do all this, we already have `axios` client set up in the `apis` folder to handle all this and also handle sending the cookies if the user is already logged in. 

Please update this to use the base client instead. Andwe'll want to also make sure we don't make the API call, if the user id is null (user is not logged in, or parent component has not yet passed in the value).

akosasante (commented):
We should also `UserProfile.js` to add a link or button on that page so that the user can easily get to this recommendations page.

src/App.js
DYA13: Akos , i was thinking that we want to have recommendations based on user logged in and his history, album purchased etc, or may be I am mistaken? Sorry I didn't catch up

src/components/RecommendationPage.js
DYA13: axios implemented , commit 3676f7a3ef5cc36ad31c1b9d1498cd46dbd188d3

src/App.js
akosasante: Yes that's correct; my previous comment was regarding getting the `user` in `App.js` to pass in a user id to the component. My suggestion was to instead have `const { user } = useAuth()` inside the `RecommendationPage` itself. But it's only a small optimization, so I'm okay with leaving it as-is.

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,45 @@
+import React, { useState, useEffect } from 'react';
+import axiosInstance from '../api/axiosClient';
```
akosasante:
```suggestion
import axiosInstance from '../apis/axiosClient';
```

src/App.js
```diff
@@ -101,6 +102,14 @@ function App() {
           }
         />
 
+        <Route
+          path="/profile/recommendations"
+          element={
+            <RequireAuth>
+              <RecommendationPage userId={user ? user.userId : null} />
```
akosasante: If you do a `console.log(user)` , you'll see that the object is actually nested and the property for the id is under `_id`:



so if we're keeping this call here instead of moving it to inside the `RecommendationPage` component, then this would need to be:

```suggestion
              <RecommendationPage userId={user ? user.user._id : null} />
```

src/App.js
DYA13: Akos , do you mean to fetch user in this way?  commit 660ec8671459cc367a4a5be452cc6da2d5b1c71f

akosasante: Yep exactly; so we should be able to remove this change here in `App.js`, and no longer need to `RecommendationPage` component on line 147

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from 'react';
+import axiosInstance from '../apis/axiosClient';
+import { Typography, CircularProgress, List, ListItem } from '@mui/material';
+import { useAuth } from '@akosasante/react-auth-context';
+const RecommendationPage = () => {
+  const { user } = useAuth();
+  const userId = user ? user.userId : null;
```
akosasante: Please take a look at my previous comment about how `user` is nested and it does not have a `userId` field. If you run `console.log` or add a breakpoint in the browser on this line, you will see what I mean.

This needs to be:

```suggestion
  const userId = (user && user.user) ? user.user._id : null;
```

or even shorter, with [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining):

```javascript
const userId = user?.user?._id
```

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,59 @@
...
+          console.error('Error fetching recommendations:', error);
+          setError('An error occurred while fetching recommendations.');
+          setLoading(false);
+        });
+    } else {
+      setLoading(false);
+    }
+  }, [userId]);
+
+  return (
```
akosasante, line 31: It'd be nice if we show some text, if we have fetched the recommendations, but the result is an empty array. Something to tell the customer "Sorry we don't have any recommendations for you yet, as you add more albums to your library we will be able to generate some" or something.

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,59 @@
...
+        <CircularProgress />
+      ) : (
+        <List>
+          {recommendations.map((album) => (
+            <ListItem key={album.albumName} sx={{ mb: 1 }}>
+              {album.albumName}
+            </ListItem>
+          ))}
+        </List>
+      )}
```
akosasante: In React, it's often a [code smell](https://martinfowler.com/bliki/CodeSmell.html) to have nested ternary operations in the render code (the code after `return` in a function component).

It is often preferred to have each state as it's own single line or statement using boolean operators instead. For example, the code here could be rewritten as:  
 

```suggestion
      {loading && <CircularProgress />}
      {error && <Typography color="error">{error}</Typography>}
      {!loading && !error && recommendations?.length === 0 && <Typography color="warning.main">Sorry no recommendations</Typography>}
      {!loading && !error && recommendations.length > 0 && (
        <List>
          {recommendations.map((album) => (
            <ListItem key={album.albumName} sx={{ mb: 1 }}>
              {album.albumName}
            </ListItem>
          ))}
        </List>
      )}
```

This works because the boolean operator `&&` checks each value to see if it's truthy, and if they are all truthy, it will return the value of the last item checked (in the above example, that would be the element we want to return in that case). 

This way when reading the code, the reader doesn't have to go through branches thinking "okay if error is false but loading is true then this element will be shown". Which can get tangled after many if/else/if/elses. Instead, each line can be read on its own: "if loading is true, then regardless of any other variables, the CircularProgress element will be shown". "if error is true, then the error message will be shown". "if error and loading are false, and the recommendations has an empty length, some text will be shown" etc.

src/components/RecommendationPage.js
```diff
@@ -0,0 +1,59 @@
...
+        fontWeight="bold"
+        color="text.secondary"
+        variant="h5"
+        spacing={0.5}
+        sx={{ mb: 2, p: 3 }}
+      >
+        Recommendations
+      </Typography>
+      {error ? (
+        <Typography color="error">{error}</Typography>
```
akosasante: The `color` attribute on the material-ui `Typography` component should map to a value from the Theme being used. 

This page shows all the values available (of course in our app some of these might have different values based on customization we've done): [https://mui.com/material-ui/customization/default-theme/#main-content](https://mui.com/material-ui/customization/default-theme/#main-content)

All the colors are under the `pallette` section. So `error` is not a color by itself, but rather an object with multiple options. So we need to pick one of those options, like for example we could make this:

```suggestion
        <Typography color="error.main">{error}</Typography>
```

and then the text will have the color "#d32f2f"

src/components/user/UserProfile.js
akosasante, line 1: 👍🏾

src/App.js
DYA13: I have removed this route in App,js commit f02e07058c3c647521957fb92caaf3a4c0347f26
Now I have eslint error that user assigned a value but never used in line 20 in App.js
`const { status, user } = useAuth();`
we used user in deleted route
Should I remove it as well?

src/App.js
DYA13: Akos in the previous commit I have deleted this route from App.js. And actually now user is not used. Should I delete it or refactor it in some way?

src/components/RecommendationPage.js
DYA13: I have removed user  in line 20 in App.js commit 94f2987a8f8c26ab2828a4c628366bde8a193ea6
May be I misunderstood regarding user 🤔, could please correct me?

I also edited variable userId in RecommendationPage.js as you proposed, thank you very much 👍
  commit  ae8502d8443ed0a13c28f9e60ddef25a7180e57a

src/components/RecommendationPage.js
DYA13: Thank you , I made  changes to show text , couls you please check? commit 0705dee8f06b5863550c25dd3f8460e13a43b671

src/components/RecommendationPage.js
DYA13: Akos I have done some changes previously in commit 0705dee8f06b5863550c25dd3f8460e13a43b671, should I change for your suggection better?

src/App.js
akosasante: We still need the route in `App.js` otherwise the app will not know what to show on: http://localhost:3000/profile/recommendations

src/App.js
akosasante: Just the `user` part can be removed. 

> , and no longer need to `RecommendationPage` component on line 147

It looks like I mistyped above ^. I did not mean that you should remove the route altogether, just remove the use of `user` attribute/prop in that route component. Apologies for the confusion.

So in the end App.js should look like:

```javascript
function App() {
  const { status } = useAuth();
  
  // etc all the existing code
  
  return (
  	// existing code, existing routes, then the recommendation one, probably right under the user profile one
  	<Route
          path="/profile/recommendations"
          element={
            <RequireAuth>
              <RecommendationPage />
            </RequireAuth>
          }
    />
  )
}
```

src/components/RecommendationPage.js
akosasante: Looks good!

src/components/RecommendationPage.js
akosasante: I think what you have now is reasonable, no need to change it :+1: . Just keep in mind for future cases, this approach for avoiding nested ternaries in render code.

akosasante (commented):
After you add back the route, in App.js this should be good to go. We will of course be able to do more debugging/revisit once we do some more work on the backend to generate recommendations.

src/App.js
DYA13: Thank you ❤️ , added, commit b7c70b6b4b564cfa8198230d5f3378742bf489c3

src/App.js
DYA13: yeah added 🙈 , thank you , commit b7c70b6b4b564cfa8198230d5f3378742bf489c3

akosasante (approved):
LGTM 🚀

PR #70, opened 2023-11-21
PS204-Firebase-store-profile-pictures

mvstoyan:


akosasante (commented):
I missed it but I think in a previous commit we accidentally deleted the route that goes to `PersonalProfileEditForm`:

6605b65#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67L35

We should add that back into `App.js` and ensure it's wrapped in the `RequireAuth` component.

Other than that, everything looks great!

mvstoyan:
> I missed it but I think in a previous commit we accidentally deleted the route that goes to `PersonalProfileEditForm`:
> 
> [6605b65](https://github.com/Code-the-Dream-School/dd-prac-team4-front/commit/6605b65a52c18970208b4fce86b132949497a41c)#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67L35
> 
> We should add that back into `App.js` and ensure it's wrapped in the `RequireAuth` component.
> 
> Other than that, everything looks great!

returned the code

src/components/user/PersonalProfileEditForm.js
```diff
@@ -9,9 +9,29 @@ import {
...
+import { Alert } from '@mui/material';
 import axiosInstance from '../../apis/axiosClient';
 import { useNavigate } from 'react-router-dom';
+
+const ProfileImage = ({ user }) => {
+  if (user && user.profileImage && user.profileImage.url) {
+    return (
+      <img
+        src={user.profileImage.url}
+        alt="user profile"
```
akosasante:
```suggestion
        alt={user.profileImage.altText || 'user profile image'}
```

akosasante (commented):
one small nit. Then LGTM. Not marking as approved for now only because I believe the agreement was to waint until all 3 image upload tickets complete; then we can compare them and pick which approach we want to merge :+1:

mvstoyan:
> one small nit. Then LGTM. Not marking as approved for now only because I believe the agreement was to waint until all 3 image upload tickets complete; then we can compare them and pick which approach we want to merge 👍

Sounds good!

akosasante (changes requested):
(no changes actually needed; just marking as such to prevent any accidental merging)

src/components/user/PersonalProfileEditForm.js
```diff
@@ -9,9 +9,29 @@ import {
...
+          height: '100px',
+          maxWidth: '100px',
+          maxHeight: '100px',
+        }}
+      />
+    );
+  }
+  return null;
+};
+
```
akosasante: Rather than using what seems to be the same code in two separate files; it would be best to export the `ProfileImage` component from the UserProfile file (or write it in its own file to be imported in both UserProfile and PersonalProfileeditForm) this allows us to reuse the same code and ensure that the visuals are consistent across components.

src/components/user/PersonalProfileEditForm.js
```diff
@@ -9,9 +9,29 @@ import {
...
-import { Avatar, Alert } from '@mui/material';
+import { Alert } from '@mui/material';
 import axiosInstance from '../../apis/axiosClient';
 import { useNavigate } from 'react-router-dom';
+
+const ProfileImage = ({ user }) => {
+  if (user && user.profileImage && user.profileImage.url) {
+    return (
+      <img
+        src={user.profileImage.url}
```
akosasante: For firebase/google cloud storage, the URLs are served from "edge nodes" (essentially google has servers all over the world so they can serve files to users super-fast by using the server that is closest to that particular user). This is great but it also means that they do a lot of caching, so sometimes when we reload the page we won't see the image right away, and will still see the old image. 

In order to get around this, we can do something called "cache-busting" where we add a random string to the URL so that Google and our browser think this is a unique URL and serve us the fresh data. 

```suggestion
        src={user.profileImage.url + `?cachebust=${Math.random()}`}
```

src/components/user/PersonalProfileEditForm.js
akosasante, line 1: To do the image upload, we'll need a few things:

1.  store the image in React state: `const [image, setImage] = useState(null);`
2.  Under the image add a file-picker (`input` html element with `type="file"` and `accept="image/*"` attributes, and an `onChange` event handelr that will get the file selected
3.  Add a button for uploading the image (separate from changing the user's name/email or password
4.  In the `onChange` handler you can get the file for the image by doing `const file = e.target.files[0]` and set the image variable in the React state to that file
5.  When the button is clicked to upload the image, you'll want to make a POST request using the `axiosInstance` to the route in your [backend PR](https://github.com/Code-the-Dream-School/dd-prac-team4-back/pull/87), if the POST request is successful you can navigate back to the `/profile` page
    1.  Because we're sending a file and not JSON, the value we pass to the axios POST function will be `new FormData()`, you can take a look at @DYA13 's PR to see how that was done there: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/72/files#diff-ae50761327201a38cc253bd05d36e6028e7596f28b198614c06044ee0d5a83deR60-R61

mvstoyan: In process, the photo is uploaded but not currently displayed.

src/components/user/UserProfile.js
```diff
@@ -23,7 +24,7 @@ export default function PersonalProfile() {
         <div className="text-center">
           <Avatar
```
akosasante: Let's replace this `Avatar` component also with the new `ProfileImage` component you've created; that way we also get the same cache-busting capabilities on this page where it will ensure that the most recently uploaded version of the image will be fetched from google's cache/storage/

akosasante (approved):
This looks great and is working well on my end, just one change needed to ensure you see the correct image on the user profile page. I'll make a separate ticket for doing a "preview" of the image selected on the edit form.

src/components/user/UserProfile.js
mvstoyan: Great, then I'll merge!

PR #71, opened 2023-11-21
Test users foto

NatachaKey:


package.json
```diff
@@ -18,6 +18,9 @@
     "@testing-library/react": "^14.0",
     "@testing-library/user-event": "^14.4",
     "axios": "^1.4.0",
+    "express": "^4.18.2",
+    "express-fileupload": "^1.4.2",
+    "mongoose": "^8.0.1",
```
akosasante: We shouldn't need these libraries in the frontend.

akosasante:
Is this PR waiting for review; or is it currently in-progress (you can mark it as a Draft if so)

NatachaKey:
@akosasante Akos, could you please take a look- i left some comment in the code

src/components/user/ImageUpload.js
```diff
@@ -0,0 +1,67 @@
...
+    formData.append('userImage', selectedFile);
+
+    try {
+      // Send a POST request to the server with the file
+      const response = await axiosInstance.post(
+        '/users/uploadUserImage',
+        formData
+      );
+
+      if (response.ok) {
```
akosasante: Axios doesn't set a `ok` variable on its response values like `fetch` does. In axios, if the response status code is not in the 200s, then it will reject the promise (which for async/await, means that an error will be raised that you can handle in the `catch` part of this try/catch). So we don't need to check if the response is ok here. 

Currently, this is causing the error message to be shown because `response.ok` is undefined, so this if/else, will go to the `else` part ofthe statement

src/components/user/ImageUpload.js
```diff
@@ -0,0 +1,67 @@
...
+        // Handle successful upload
+        setUploadStatus('Success');
+        // Refresh the page
+        // AKOS: this one below  doesn't work
+         // a delay before reloading the page 
+      // setTimeout(() => {
+      //   window.location.reload();
+      // }, 1000);
```
akosasante: Once the above is fixed, this should work. 

This refresh is fine in this case in my opinion but I do want to note that it's a little against usual React/Single-Page-App patterns, where we usually want to avoid refreshing the page unless needed. If what we need is to reload a piece of data needed on the page, passing in a function from the parent component to tell it to reload the user data would be more idiomatic.

src/components/user/ImageUpload.js
```diff
@@ -0,0 +1,67 @@
...
+      // setTimeout(() => {
+      //   window.location.reload();
+      // }, 1000); 
+
+      } else {
+        // Handle errors
+        setUploadStatus('Error');
+      }
+    } catch (error) {
+      //AKOS: could you please take a look why does it throw an error if the img is being stored and when manually refresh the page it is shown correctly
```
akosasante, line 40: See above comment

akosasante (commented):
Is it possible to remove the non-image-upload related changes in this PR?

akosasante (changes requested):
I think we can close this PR for now, if the group wants to proceed with the Firebase approach :) 

I haven't read the other open PRs yet but it looks like this one also includes some AlbumReview and styling changes so those you may want to copy into another branch. You can do that by following the following steps:

```
// start from main with the fresh set of commits

git checkout main

git pull

git checkout -b <new branch name>

git checkout testUsersFoto src/components/album/Album.js <and any other file paths that you want to copy into the new branch>
```

NatachaKey:
@akosasante Thank you so much, Akos! I´ve moved necessary  files to a new branch and opened a new PR where I will try to fix the review's functionalities
UPD: closed new PR, because it´s working version is in the branch `cartPageCreated` , PR:  https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/81

PR #72, opened 2023-11-22
When registering, user should be able to upload a profile picture PS-205

DYA13:


DYA13:
I have started to create function for uploading picture. It is not finished yet. Could you please check if I have started correctly?

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,11 +45,21 @@ export default function PersonalProfileEditForm() {
...
-          name: formData.name,
-          email: formData.email,
-        });
+        const response = await axiosInstance.patch(
+          '/users/updateCurrentUser',
+          {
+            name: formData.name,
+            email: formData.email,
+          },
+          formDataForApi
```
akosasante: You can't add the `formDataForApi` here; unless you're also updating the `updateCurrentUser` endpoint to handle these values.

If you will be making the call to the code that's currently in your backend `imgur` branch, then this should make two separate API calls, one to the `updateCurrentUser` to update the name and email. And then one to `/users/:userId/uploadProfile` where you would send the profile image via form data.

akosasante (commented):
Left a comment about making the API request, otherwise looks good so far

src/components/user/PersonalProfileEditForm.js
DYA13: I removed formDataForApi  , commit 59642fa0e083a3bb478c57b3ecd24ebd46e2b2ce

DYA13: Now we have functionality in this way that we can't change only photo, we need to change name and email. What if the user wants to change only photo,we don't have this opportunity to give for user. What do you think , should add this functionality? 
image

akosasante: Yeah I think it's reasonable for the user to only want to change their picture. I can think of a few approaches:

1.  Separate form+button for "Save Profile Photo" that only handles making the request to save the photo. And the existing "Save Changes" button can handle just saving the name/email change.
2.  Populate the name and email inputs with the existing name and email to start with. That way, when we click "Save Changes" those will just send an update request to keep them the same, while the profile update request goes out
3.  In the code, check if _both_ name and email are empty, if they are skip making that request altogether.

DYA13: I tried to apply the second option, commit ef472280e946654d8b22b4e74fabf05a0f80d068
Could you please check?

src/components/user/PersonalProfileEditForm.js
```diff
@@ -147,7 +163,7 @@ export default function PersonalProfileEditForm() {
                 <TableCell>
                   <TextField
                     name="name"
-                    value={formData.name}
+                    value={formData.name || userData.username}
```
akosasante: Making this change here will lead to some issues; because now we are setting the value of the text field; but back in our `handleProfileSubmit` function were are setting errors if `formData` is empty:

```
      setErrors({
        name: !formData.name ? 'Name is required' : '',
        email: !formData.email ? 'Email is required' : '',
      });
```



Your first thought might be to add a check in `setErrors` that also does the same `!formData.name || !userData.name` but I feel that starts to get complicated if we need to be updating multiple spots to do this kind of conditional check.

Instead, I would rather set the `formData` when we fist render this component to the values that we want it to be.

Setting values when the component first renders, can be done with a `useEffect` hook:

```javascript
  useEffect(() => {
    setFormData({
      name: userData?.username,
      email: userData?.email,
      oldPassword: '',
      newPassword: '',
      profilePicture: null,
    });
  }, [])
```

src/components/user/PersonalProfileEditForm.js
```diff
@@ -160,7 +176,7 @@ export default function PersonalProfileEditForm() {
                   <TextField
                     name="email"
                     type="email"
-                    value={formData.email}
+                    value={formData.email || userData.email}
```
akosasante: Same note as above comment

src/components/user/PersonalProfileEditForm.js
akosasante, line 1: This component is missing code inside `handleProfileSubmit` to update the profile image. We currently are calling `patch('/users/updateCurrentUser',...)` to update the user's name and email; but we need to also handle calling the endpoint you've created on the backend to upload the image to imgur.

src/components/user/PersonalProfileEditForm.js
DYA13: I have added useEffect , I added as well userData as a dependency to rerun the effect when userData changes,I am not sure if I did correctly🫣
commit 8a1154cd88df9d49ec24287aa8319ee8c1808e60

src/components/user/PersonalProfileEditForm.js
DYA13: I added code in  handleProfileSubmit function , could you please check?
commit 42cb65cfbd1173ba7ed32f90b08c52c6e0ec3431

DYA13:
There is an error with userData , I didn't find yet but from this branch I don't manage to sign in and log in as well🤔

image

akosasante:
> There is an error with userData , I didn't find yet but from this branch I don't manage to sign in and log in as well🤔
> 
> image

Please try rebasing your branch on main (`git rebase origin main`) and resolving any conflicts that arise. Then push that to Github (you may need to add a flag if there were conflicts you fixed. Like this: `git push --force-with-lease`).

Those are all pointing to these lines in the `PersonalProfileEditForm.js` and saying that the `userData` variable is not defined. Looking at that file; it seems this stems from a change from a litlte while ago (sorry, I made this and posted a little late in Slack that a rebase would be needed!) that change the variable on line 30 to be `user` instead of `userData` since we no longer need to do the nested access (user.user.whatever) that we had to do before: [https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/src/components/user/PersonalProfileEditForm.js#L29](https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/main/src/components/user/PersonalProfileEditForm.js#L29)

Maybe this got mixed up during a previous pull from main or rebase on main?

To fix it; all the places in this file that are using `userData` you can rename those variables to use `user` instead.

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,6 +55,23 @@ export default function PersonalProfileEditForm() {
...
     } else {
       try {
+        // If there's a profile picture, upload it first
+        if (formData.profilePicture) {
+          const formDataImage = new FormData();
+          formDataImage.append('image', formData.profilePicture);
+
+          // Make the API call to upload the image
+          const imageResponse = await axiosInstance.post(
+            '/:userId/uploadProfile',
```
akosasante: This will actually make a call to `/:userId/uploadProfile` axios will assume that is the literal string you want to use:



To substitute a variable we can use Javascript Template Strings (the backticks \` \` ). You are also missing the `/users` part of the url (remember that `axiosInstance` will handle the base part of the url (<localhost:8000 or musicstore.onrender.com>/api/v1`) but we still need to provide the rest: 

```javascript
          const imageResponse = await axiosInstance.post(
            `/users/${user?._id}/uploadProfile`,
            formDataImage
          );
```

src/components/user/PersonalProfileEditForm.js
akosasante, line 1: We should also update the code in `UserProfile.js` to show the uploaded profile image in the avatar component

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,6 +55,23 @@ export default function PersonalProfileEditForm() {
       });
     } else {
       try {
+        // If there's a profile picture, upload it first
+        if (formData.profilePicture) {
+          const formDataImage = new FormData();
+          formDataImage.append('image', formData.profilePicture);
```
akosasante: The backend is expecting that we send this file with the name `profile` not `"image"`

```suggestion
          formDataImage.append('profile', formData.profilePicture);
```

Here: https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/311fd8c5230019e23d6933c28811a08f81566a89/src/controllers/imgurController.js#L35-L41

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,6 +55,23 @@ export default function PersonalProfileEditForm() {
...
+          const imageResponse = await axiosInstance.post(
+            '/:userId/uploadProfile',
+            formDataImage
+          );
+
+          // Assuming the backend returns the image URL, update the formData
+          setFormData((prevData) => ({
+            ...prevData,
+            profilePicture: imageResponse.data.imageUrl,
+          }));
```
akosasante: I think we can remove this code. We are already displaying the image when the user selects it from the file; so there's nothing to change here.

In addition, this will currently break things:



This error shows up after I successfully upload a profile image. This happens because `imageResponse.data.imageUrl` is a String; but down on line 157, we are calling `URL.createObjectUrl(formData.profilePicture)`. `createObjectURL` is a function that expects to receive a **file** (in javascript called a Blob) but now we're gonna be passing it a String. 

Let's avoid this by just removing these lines altogether for simplicity.

akosasante (commented):
Looking good; a few remaining things to tweak and then should be great!

Extra things we can consider in a later ticket:

*   loading the existing image when the profile edit page first loads
*   Refreshing the user object before navigating back to the user profile page, so that the latest updates are available

src/components/user/PersonalProfileEditForm.js
DYA13: Thank you , changed commit f6aad6f76ab1b16a05fd7af3c08e91400292675f

src/components/user/PersonalProfileEditForm.js
DYA13: could you please check if it is correct, commit 957416af5a5747c660c0fbbb6ba015c937f0b5af

src/components/user/PersonalProfileEditForm.js
DYA13: thank you ❤️

src/components/user/PersonalProfileEditForm.js
DYA13: I have removed this function , commit ba8b8cdb428068919638872c1554313871677e58
Now the variable imageResponse has eslint error

DYA13:
Now when I want to save image , I have errors with user data 🙈
image

src/components/user/PersonalProfileEditForm.js
```diff
@@ -29,6 +30,16 @@ export default function PersonalProfileEditForm() {
   const { user } = useAuth();
   const userData = user.user; //he user that's returned is nested in its original response shape so to use the actual user you'll need to unwrap it
 
+  useEffect(() => {
+    setFormData({
+      name: user?.username || '',
+      email: user?.email || '',
```
akosasante: Above on line 31, we store the user info in the `userData` variable.

But here you're trying to get `user?.username` which will turn out to be undefined. You can see this if you add a console log inside the useEffect, right between line 33 and line 34:





That is why you're seeing those errors because when we click "Save Changes" button, we try to submit the current value of the `formData`  variable. But we first check and set errors on line 50-54 if the `formData.name` or `formData.email` fields are empty.  Currently they are empty because we're not setting the values correctly at first.

If you change these lines to:

```suggestion
      name: userData?.username || '',
      email: userData?.email || '',
```

that will avoid that particular issue

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,6 +55,17 @@ export default function PersonalProfileEditForm() {
...
     } else {
       try {
+        // If there's a profile picture, upload it first
+        if (formData.profilePicture) {
+          const formDataImage = new FormData();
+          formDataImage.append('profile', formData.profilePicture);
+
+          // Make the API call to upload the image
+          const imageResponse = await axiosInstance.post(
+            `/users/${user?._id}/uploadProfile`,
```
akosasante: Same as above; you'll need to change this to refer to `userData`

```suggestion
            `/users/${userData?._id}/uploadProfile`,
```

src/components/user/PersonalProfileEditForm.js
```diff
@@ -44,6 +55,17 @@ export default function PersonalProfileEditForm() {
       });
     } else {
       try {
+        // If there's a profile picture, upload it first
+        if (formData.profilePicture) {
+          const formDataImage = new FormData();
+          formDataImage.append('profile', formData.profilePicture);
+
+          // Make the API call to upload the image
+          const imageResponse = await axiosInstance.post(
```
akosasante: The linter warns us that there is no need to assign the result to a variable if we're not going to use the variable. We can make the linter happy by just adding an underscore at the front of the variable:

```suggestion
          const _imageResponse = await axiosInstance.post(
```

or we can remove the variable altogether (the axios POST function would still be called, we're just not storing the value anywhere):

```suggestion
          await axiosInstance.post(
```

akosasante (changes requested):
From the Slack channel, it sounds like we can go forward with the Firebase approach for uploading profile image; but still wanted to leave some last comments here to outline how to fix the error you were seeing. Great work!

src/components/user/PersonalProfileEditForm.js
DYA13: Thank you very much❤️ Have I understood correctly that  we obtain userData from useAuth hook here?

src/components/user/PersonalProfileEditForm.js
DYA13: Thank you ❤️

src/components/user/PersonalProfileEditForm.js
DYA13: Thank you so much❤️, I didn't know that by adding underscore at the front of the variable, we can resolve this issue

src/components/user/PersonalProfileEditForm.js
DYA13: Akos , 
I have made the changes according to our chat above  but I still have errors, this time about axios request. 
image

DYA13:
Akos , 
I have made the changes according to our chat above  but I still have the errors, this time about axios request. 
image

akosasante:
@DYA13 -- hmm I'm not seeing the same on my end; you're running this with the `imgur` branch on the backend right? Can you expand the `request` section in that AxiosError to verify what URL it was calling where it got a 404? What does it show in the Chrome Network tab for the request with the 404? How about in the backend server logs?

DYA13:
> @DYA13 -- hmm I'm not seeing the same on my end; you're running this with the `imgur` branch on the backend right? Can you expand the `request` section in that AxiosError to verify what URL it was calling where it got a 404? What does it show in the Chrome Network tab for the request with the 404? How about in the backend server logs?


I have pulled latest changes merged from 'main' on front today, I have more console errors 
Before merging the changes, I was running with 'main' branch, not 'imgur' branch. 
Now I don't manage to sign in or sign up . 
image

image

image
 
All errors 
image

In network I don't see any additional info about errors
Backends logs errors 
image

akosasante:
@DYA13 Hmm that's odd.

Do you have issues if you run `main` branch on both the frontend and backend?

What about with `imgur` on the backend and this branch in the frontend?

I would like to see the network requests in the "Network" tab when you're seeing these errors, that will provide a bit more info about what status/errors the HTTP requests are getting rather than just the Axios errors in the console. 

What logs do you see in the backend when trying to log in? In your screenshots I'm only seeing the requests for fetching a wishlist or fetching albums but I'm not seeing the login or sign-up requests. 

If you want to chat about this together on zoom feel free to take a look and see if there's a time on my calendar that works for you: https://www.when2meet.com/?25229680-wSzu8

PR #73, opened 2023-11-28
 error deleting the review, expects reviewId from req.params

NatachaKey:


src/components/review/ReviewsList.js
```diff
@@ -77,6 +78,9 @@ const AlbumReviews = ({ albumId }) => {
           reviewId={reviews[0]?._id}
         />
       )}
+       {user && userHasReviewed && (
+          <DeleteReview reviewId={reviews[0]?._id} refreshReviews={refreshReviews} />
```
akosasante: Hmm, how do you know the current user's review will always be the first one in the `reviews` list?

akosasante (commented):
While testing out this code, I found a small bug in the backend. Do you think you'll have a chance to take a look at fixing it?:

https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/e9bfbdac6499460dfbd35c7e23c2164f5776b0db/src/controllers/reviewController.js#L140

`CustomError.ForbiddenError` does not exist. We should either be using `CustomError.UnauthorizedError` or directly returning a status with `StatusCodes.FORBIDDEN`.

src/components/review/ReviewsList.js
NatachaKey: @akosasante just changed this error status on the backend ++ I was asking myself the same thing about passing this  `{reviews[0]?._id} ` as a prop, now added code where use a find method- does it work properly? // also I am thinking maybe I shoul place this 'DELETE' button in another place..

src/components/review/ReviewsList.js
akosasante, line 1: One thing to note here; the code you have here will work; but one note that comes up in a lot of the React information I read is to try to reduce the amount of `useState` variables in a given component. This reduces the amount of times that React will need to re-render a component due to a state variable changing. It also simplifies things as the component grows and gets more complex.

Specifically, in this case, I am looking at the `userHasReviewed` state variable. 

Inside `fetchAlbumReviews`, we are calling `setReviews` to update the list of reviews in the component state. This will trigger a component re-render. Then we also call `setUserHasReviewed`, which will again trigger a component re-render. And then lastly we call `setLoading` which triggers a component re-render. Now, React is advanced enough to see that those are all happening in the same function and will try to batch those actions so only one re-render will occur. But we can still help React out by not using Set State when it's not needed.

In this case, the reason I say it's not needed is because we can _derive_ (ie: calculate) the value of `userHasReviewed` based on the value of an existing state variable: `reviews`. Once `reviews` is set, then we can figure out `userHasReviewed` without needing to explicitly set state.

This would look like:

```javascript
  const [reviews, setReviews] = useState([]);
  const [loading, setLoading] = useState(true);
  const [error, setError] = useState(null);
  const { user } = useAuth();
  // whenever the component re-renders, this line will be read by React
  // If `reviews` is empty or null, then this will return false
  // Otherwise, this will enter the `some` and check if any of the reviews have a matching user id
  // This is a regular variable that gets its value based on doing some calculation on a state vairable
  // But since it's a normal variable and not a state variable itself, we don't need to call `set state` on it and the componsnet doesn't need to re-render to store this result 
  const userHasReviewed = (reviews || []).some((review) => review.user === user?._id);

  const fetchAlbumReviews = useCallback(async () => {
    try {
      const response = await axiosInstance.get(`/reviews/album/${albumId}`);
      console.log(response.data.allProductReviews);

      const { allProductReviews } = response.data;
      setReviews(allProductReviews);
      setLoading(false);
    } catch (error) {
      console.error('Error fetching album reviews:', error);
      setError('Error fetching album reviews. Please try again later.');
      setLoading(false);
    }
  }, [albumId, user?._id]);
```

src/components/review/ReviewsList.js
```diff
@@ -67,10 +68,20 @@ const AlbumReviews = ({ albumId }) => {
...
       )}
+
       {/* Conditionally render the WriteReview component */}
       {user && !userHasReviewed && (
-        <WriteReview albumId={albumId} refreshReviews={refreshReviews} />
+        <WriteReview
+          albumId={albumId}
+          refreshReviews={refreshReviews}
+          userHasReviewed={userHasReviewed}
+          reviewId={reviews[0]?._id}
```
akosasante: Currently, the `WriteReview` component doesn't seem to accept these props (`userHasReviewed` and `reviewId`). Can they be removed?

src/components/review/ReviewsList.js
```diff
@@ -67,10 +68,20 @@ const AlbumReviews = ({ albumId }) => {
...
+        />
       )}
+       {user && userHasReviewed && (
+          <DeleteReview reviewId={
+            reviews.find((review) => review.user === user?.user?._id)?._id} refreshReviews={refreshReviews} />
```
akosasante: I've updated `useAuth` in `main` branch to no longer need this nested-access (😮‍💨). So now we can simplify accessing the user to:

```suggestion
            reviews.find((review) => review.user === user?._id)?._id} refreshReviews={refreshReviews} />
```

src/components/review/DeleteReview.js
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import axiosInstance from '../../apis/axiosClient';
+//import { useAuth } from '@akosasante/react-auth-context';
```
akosasante: Not needed?

src/components/review/DeleteReview.js
```diff
@@ -0,0 +1,40 @@
...
+
+const DeleteReview = ({ reviewId, refreshReviews }) => {
+  const [deleting, setDeleting] = useState(false);
+  const [error, setError] = useState(null);
+  // const { user } = useAuth();
```
akosasante: can be removed?

akosasante (commented):
In terms of _where_ to put the Delete button, I think where it is now is fine. 

If you wanted you could move it to the place in the code where we're iterating over the reviews to show them each. This would have an advantage of being able to show the Delete button next to the review that belongs to that user. As well, you then would not need to do a `.find` because you would already have the current `review` you're mapping over in scope; and can instead just pass `review._id` into the `DeleteReview` component.

NatachaKey:
@akosasante  UPD: resolved in another PR. 

old msg: Akos could you please take a look when you have a minute- I've tried to place delete btn below the user's unique review to be deleted but not sure why i don't see it al all-

NatachaKey:
I think this PR is covered in https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/74/

akosasante (approved):
Looks good! As you said; you have the fix for the `Delete` button showing up under all the comments in #74 .

You can either in Github merge that PR into this one (I think you already have the branches set up that way); then if you need to do any further cleanup you can do `git checkout deleteReview; git pull origin deleteReview` locally; make any changes then push to this branch. Then go ahead and merge once you're confident all the changes together are working as expected.

akosasante:
One thing I wanted to add; you've been using plain `<button>` tags which will use the unstyled buttons provided by the browser. We can take advantage of the theming and styling that we have available to us from `material-ui` by instead using the `<Button>` component available to import from `import { Button } from '@mui/material` . For the most part, the button should work the same anyway. The docs for the different attributes available on the Button component are here: https://mui.com/material-ui/react-button/

PR #74, opened 2023-11-29
edit review option added on the frontend

NatachaKey:


src/components/review/ReviewsList.js
akosasante, line 1: Per the githubActions lint warning for line 19; this dependency is not needed. We should only be listing variables or functions that are used _inside_ the `useEffect` as dependencies. And since `fetchAlbumReviews` is the only thing we have inside; that is the dependency we should be listing, rather than `albumId`. In order to do that, we need to ensure `fetchAlbumReviews` is a variable rather than a function; which requires wrapping it in a `useCallback` hook. https://react.dev/reference/react/useEffect#specifying-reactive-dependencies

src/components/review/ReviewsList.js
```diff
@@ -80,13 +80,21 @@ const AlbumReviews = ({ albumId }) => {
...
-          <DeleteReview reviewId={
-            reviews.find((review) => review.user === user?.user?._id)?._id} refreshReviews={refreshReviews} />
-        )}
+        <UpdateReview
+          reviewId={
+            reviews.find((review) => review.user === user?.user?._id)?._id
+          }
+          refreshReviews={refreshReviews}
+        />
+      )}
```
akosasante: This works well 👍🏾👍🏾

One improvement we could try is instead making it so we can click on our review to edit it; and only then show this `UpdateReview` component. 

https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/271b8dfe-a4bd-4d3a-af87-8789da200c43

akosasante: As a starter; this would require moving the `UpdateReview` to inside the `reviews.map` section, similar to the suggestion I gave in the delete review PR: https://github.com/Code-the-Dream-School/dd-prac-team4-front/pull/73#pullrequestreview-1765258503

akosasante (commented):
Looking great!

src/components/review/ReviewsList.js
NatachaKey: @akosasante  I think I did it! check it out (as well as a delete btn)

NatachaKey:
@akosasante I think this is it=) Akos, please check the functonality

src/components/review/ReviewsList.js
```diff
@@ -28,8 +28,7 @@ const AlbumReviews = ({ albumId }) => {
       setError('Error fetching album reviews. Please try again later.');
       setLoading(false);
     }
-  }, [albumId, user?._id]);
-
+  }, [albumId, user?.user?._id]);
```
akosasante: Per the linter; we are not using `user.user.id` inside the `useCallback` so we can go ahead and remove it so that React is not unecessarily tracking that dependency

```suggestion
  }, [albumId]);
```

src/components/review/ReviewsList.js
```diff
@@ -4,17 +4,17 @@ import WriteReview from './WriteReview';
...
 import DeleteReview from './DeleteReview';
+import UpdateReview from './EditReview';
 
 const AlbumReviews = ({ albumId }) => {
   const [reviews, setReviews] = useState([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState(null);
-
-  const { user } = useAuth(); //use user.user.<whatever field we want> to access it properly
+  const [updateClicked, setUpdateClicked] = useState(false);
```
akosasante, line 13: Just a nitpick; I would probably call this something like `isEditing` instead of `updateClicked` but that's just a wording preference

akosasante (approved):
Great work! Left some small notes but overall looks good to me!

Follow up work that could be added as separate tickets if desired:

*   a dialog that pops up before deleting that asks if you're sure you want to delite this comment
*   A cancel button to have the edit form go away (video below)
*   Ensure that the error message goes away once the user fixes their issue and potentially show the error message returned by the server for 4xx status codes so that the user has more info about what they need to fix

akosasante:
https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/fffc568e-b663-4aec-8716-071f10fb9056


https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/50638276-ccc9-49a7-8cfe-de0a1d194395

PR #75, opened 2023-11-29
On a user's profile page, show what album they are currently listening to PS-209

DYA13:


DYA13:
Akos , could you please instruct/suggest how I can proceed with the ticket? Thank you❤️

akosasante:
This is a good opportunity to practice websockets on both the React frontend and in the backend.

On the backend, we already have socket.io events that are emitted when a user joins an album listening room and plays/pauses on that page:

[https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/live/index.js#L30-L52](https://github.com/Code-the-Dream-School/dd-prac-team4-back/blob/main/src/live/index.js#L30-L52)

Let's extend on that a bit now. Since we only really care about what a user is _currently_ listening to, we don't necessarily need to save that information more permanently in the database. For now, it would be enough to just store that info in-memory in the server. We can do that by just creating an object in the `live` folder that looks like:

```javascript
const currentlyListenedToAlbums = {}; // Key: userId, Value: albumId
```

*   When a `listening-to-album-play` event happens, add the userid and album id to the `currentlyListenedToAlbums` object
    *   Emit a message to the `"user_listening:${userId}"` room indicating that the user is listening to this `albumId`
*   When a `listening-to-album-pause` event or `user-leaving-page` event happens, remove that userId/albumId pair from the `currentlyListenedToAlbums` object
    *   Emit a message to the `"user_listening:${userId}"` room indicating that the user is no longer listening to an album
*   Create a new socket event `join:currently-listening`, that joins that socket user to a room  (roomName = `"user_listening:${userId}"`
    *   [https://socket.io/docs/v3/rooms/](https://socket.io/docs/v3/rooms/)
    *   [https://stackoverflow.com/questions/34909323/socket-io-how-to-correctly-join-and-leave-rooms](https://stackoverflow.com/questions/34909323/socket-io-how-to-correctly-join-and-leave-rooms)
    *   It should also emit the current albumId for that user id from the `currentlyListenedToAlbums` object

In the frontend:

*   On the user profile page, have a state variable for the currently listened to album, and display it somewhere on the page
*   On the user profile page, when the page loads, set up a socket and emit a message to the new socket `join:currently-listening` , use the return value to update the currently listened to album
*   Also set up a socket that listens to events to the `user_listening:${userId}` room, for any changes to the currently listened to track

Let me know if you have any qs of how to get started. In the frontend you can refer to the `AlbumChatBox` component for references.

PR #76, opened 2023-12-04
Use PropTypes library to document <Navbar> and auth components PS-156

DYA13:


src/components/Navbar.js
```diff
@@ -135,5 +136,15 @@ const Navbar = ({ toggleDarkMode, mode }) => {
...
+  toggleDarkMode: PropTypes.func.isRequired,
+  mode: PropTypes.string.isRequired,
+  handleClearCart: PropTypes.func.isRequired,
+  isLoggedIn: PropTypes.bool,
+  statusOptions: PropTypes.oneOf(['active', 'inactive']),
+  userDetails: PropTypes.shape({
+    username: PropTypes.string,
+    email: PropTypes.string,
+  }),
+};
```
akosasante: `propTypes` is only used for documenting the `props` of a component. For function components, that means, the arguments in the function definition `const Navbar ({ toggleDarkMode, mode}) => ...` . So we can remove anything that is not a prop; otherwise, right now we get warnings like this in the browser console:



Unfortunately, there's not really a straightforward way of documenting internal variables of a component, or state variables, apart from using JSDoc (similar to how we document our API specs in the backend) or Typescript. 

Additionally, we can make `mode` even more specific by using the `propTypes.oneOf` function which allows us to tell the prop checker that this value can only be a specific set of values (in this case, we expect mode to be either `dark` or `light` and would want to get warned if it somehow was being set to anything else)

```suggestion
Navbar.propTypes = {
  toggleDarkMode: PropTypes.func.isRequired,
  mode: PropTypes.oneOf(['light', 'dark']).isRequired,
};
```

And example of how this works; I set mine to `oneOf(['dark', 'medium'])` and it showed the following warning in the browser console:

src/components/userAuth/Logout.js
```diff
@@ -32,5 +33,17 @@ const Logout = () => {
...
+  errorHandler: PropTypes.func,
+  submit: PropTypes.func,
+  navigate: PropTypes.func,
+  signOut: PropTypes.func,
+  status: PropTypes.string,
+  statusText: PropTypes.string,
+  user: PropTypes.object,
+  logout: PropTypes.func,
+  handleLogout: PropTypes.func,
+};
```
akosasante: Since this component is not expecting any props; we should not need to add propTypes here after all. For now at least :smile:

src/components/userAuth/SignIn.js
```diff
@@ -166,3 +167,15 @@ export default function SignIn() {
...
+  apiUrl: PropTypes.string,
+  errorHandler: PropTypes.func,
+  submit: PropTypes.func,
+  navigate: PropTypes.func,
+  signOut: PropTypes.func,
+  status: PropTypes.string,
+  handleChange: PropTypes.func,
+  handleLoginError: PropTypes.func,
+  handleSubmit: PropTypes.func,
+};
```
akosasante: Same with this component; we can remove all the changes in this file since it's not expecting any props.

akosasante:
https://www.youtube.com/watch?v=eH7Vxmjfvf0&t=26s

src/components/userAuth/Logout.js
DYA13: thank you Akos ❤️ , commit 1bd5ab26f63ad7a21b158e0a26a812f1b339021e

src/components/userAuth/SignIn.js
DYA13: Thank you , commit a180c241a567fc4777440760cc7d1d8f38f1a5b3

akosasante (approved):
:ship:

PR #77, opened 2023-12-04
Add loading and transitions PS-171

DYA13:


DYA13:
Akos , I have done loading for SignIn. Could you please check if I am on the right way? Could you advice where I can apply more loading features and where I can implement transitions? May be I can use some libraries or my method is ok ? Thank you very much

src/components/userAuth/SignIn.js
```diff
@@ -21,6 +21,7 @@ export default function SignIn() {
     email: '',
     password: '',
   });
+  const [loadingSubmit, setLoadingSubmit] = useState(false);
```
akosasante, line 25: :+1: 

Looks good; but where are we calling `setLoadingSubmit` to enable the button change?

akosasante (commented):
This ticket was left open-ended intentionally to allow some exploration. I did mean to add some guidance and resources but hadn't gotten to it yet 😓

There's a number of places where we're making API calls where we could add this kind of loading state as you've started here. We could also consider using Material-UI's loading button: [https://mui.com/material-ui/react-button/#loading-button](https://mui.com/material-ui/react-button/#loading-button) where appropriate.

For other ideas/things to look into:

*   [react-transition-group library](https://reactcommunity.org/react-transition-group/switch-transition) (it doesn't animate on its own but it's worth learning a bit about since it's used in a bunch of other places)
	* https://blog.logrocket.com/react-transition-group/
	* https://www.youtube.com/watch?v=sfZZA8n07FA 
*   transition components given by material-ui let us do things like fading in individual components. for example: [https://mui.com/material-ui/transitions/](https://mui.com/material-ui/transitions/)

https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/09590110-3c2e-4272-9246-b72077e0a4c0

* We can use `Suspense` component from React itself to fallback to loading states when an API is loading instead of having to set state variables ourselves
	* https://react.dev/reference/react/Suspense#displaying-a-fallback-while-content-is-loading	
	* https://www.youtube.com/watch?v=L6gG1Y_T2Cs

* We can also explore using the Framer Motion library for page transitions: https://www.framer.com/motion/examples/#react-router-6-page-transitions
	* 	https://www.youtube.com/watch?v=8voRA_DMuFM&pp=ygUlZnJhbWVyIG1vdGlvbiByZWFjdCBwYWdlIHRyYW5zaXRpb25zIA%3D%3D
	* https://www.youtube.com/watch?v=FdrEjwymzdY&pp=ygUlZnJhbWVyIG1vdGlvbiByZWFjdCBwYWdlIHRyYW5zaXRpb25zIA%3D%3D

src/components/userAuth/SignIn.js
DYA13: yeah it is true , we are not calling setLoadingSubmit . Is there is any specific function provided for controlling the loading state in useLogin hook ?🤔

akosasante (approved):
I think this PR is looking great, I will write up some tickets for further transitions and work to generalize this loading pattern you've implemented here further.

PR #78, opened 2023-12-05
PropTypes home/app

mvstoyan:


mvstoyan:
Do we need to add PropTypes for Home.js and App.js here? In my understanding, it's not necessary?

akosasante (changes requested):
Oh shoot; yeah you're right I don't know how I missed that when writing the ticket :facepalm: 

None of these three components need any `propTypes` documentation, because right now, none of them take any props at all. 

We can just close this PR; sorry for the wasted work!

akosasante (approved):
:ship:

PR #79, opened 2023-12-07
first socket bug commit

NatachaKey:


src/App.js
```diff
@@ -158,6 +159,14 @@ function App() {
...
           }
         />
+        <Route
+          path="/listening"
+          element={
+            <RequireAuth>
+              <ListeningAlbum />
+            </RequireAuth>
+          }
+        />
```
akosasante: To clarify, is this route for testing purposes or do we expect to also use it for real as a separate page users can navigate to?

I ask because right now if I navigate to `http://localhost:3000/listening`, this component will be loaded but it won't get any album id passed to it. We could pass in an album via the URL (eg: `http://localhost:3000/listening/:albumId` ) for example.

akosasante: Oh I see your comment about this at the bottom of `ListeningAlbum.js`now. :sweat\_smile: 

So yes, we could define a route like this: https://reactrouter.com/en/main/hooks/use-params

```javascript
        <Route
          path="/listening/:albumId"
          element={
            <RequireAuth>
              <ListeningAlbum />
            </RequireAuth>
          }
        />
```

And then we can grab that `albumId` value in the component with `useParams()` function from `react-router-dom`. 

So one question I have about your approach so far; you have the `<ListeningAlbum>` component available in the `Album.js` modal component as well. There you are passing in the `albumId` via props. Do we want to have both approaches ? I think that's fine we just have to decide which would take precedence if there's ever a case where the component is passed an album id via the URL and also via props.

src/components/album/ListeningAlbum.js
akosasante, line 1: Right now this code _does_ work; but playing around with it shows that we may want to adjust some things on the backend. Main thing I notice is that we only `join` a listening room, by first sending an event to the _play_ or _pause_ channels. Depending on how we want this to work; this might be okay. Do we want users to only be able to see other users who are listening ot the same album _after_ the user first clicks "play" ? Do we want to give them the ability to stop receiving events about others listening to an album? Of course, the socket currently will disconnect once this component unmounts, so at that point they will stop receiving messages (unless we want that to work differently).


https://github.com/Code-the-Dream-School/dd-prac-team4-front/assets/9094098/04251670-77f1-43ef-8721-47e0a6c7aff7

src/components/album/ListeningAlbum.js
```diff
@@ -0,0 +1,70 @@
...
+    document
+      .getElementById('playButton')
+      ?.addEventListener('click', handlePlayClick);
+    document
+      .getElementById('pauseButton')
+      ?.addEventListener('click', handlePauseClick);
+    window.addEventListener('beforeunload', handleBeforeUnload);
+
+    // Cleanup event listeners when the component unmounts
+    return () => {
```
akosasante: We should also be cleaning up the `socket` connection when the component unmounts, to prevent memory leakage. You can refer to the similar code in `AlbumChatBox.js` where this is done: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/ef4ddbe8907bc63366ab62125db37e5524cac3b0/src/components/album/AlbumChatBox.js#L71-L77

src/components/album/ListeningAlbum.js
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    // Add event listeners when the component mounts
+    document
+      .getElementById('playButton')
+      ?.addEventListener('click', handlePlayClick);
+    document
+      .getElementById('pauseButton')
+      ?.addEventListener('click', handlePauseClick);
+    window.addEventListener('beforeunload', handleBeforeUnload);
```
akosasante: While these will work in this case since the component is fairly simple; it may become problematic as we expand on the functionality. Generally, it is best practice to _not_ use `addEventListener` in React code. This is because those interact with the DOM directly. Which means when React is re-rendering this component (for example if a state variable or path variable changes) then it may be re-adding these event listeners in duplicate. In this case it's unlikely since this `useEffect` is only rerun if the `user id` or `albumId` change. But it's still best to stick to React's own listeners when possible. 

https://legacy.reactjs.org/docs/handling-events.html#:~:text=When%20using%20React%2C%20you%20generally%20don%E2%80%99t%20need%20to%20call%20addEventListener%20to%20add%20listeners%20to%20a%20DOM%20element%20after%20it%20is%20created.%20Instead%2C%20just%20provide%20a%20listener%20when%20the%20element%20is%20initially%20rendered.

For `click` listeners, this would mean adding `onClick` attribute to the buttons. Eg:

```javascript
// move handler to outside of the `useEffect`

// Function to handle play button click
const handlePlayClick = () => {
  console.log('playing')
  const userId = user?._id;
  socket.emit('listening-to-album-play', { albumId, userId });
};

// In the render (return statement) tell React to call the handler when the button element is clicked on
// Render the component
return (
  <>
    <h1>Listening to Album</h1>
    <button onClick={handlePlayClick} id="playButton">Play</button>
    <button id="pauseButton">Pause</button>
  </>
);
```

Same can be done for the pause button. Since these functions need access to the `socket` variable and we want to make sure we're using the same `socket` regardless of if React needs to re-render this component, we should store that in a `useState` variable. You can refer to `AlbumChatBox.js` for the approach used there.

For `handleBeforeUnload`, it sort of works. It's listening for a `beforeunload` DOM event. So if the other user closes their tab or window; or does a full refresh of the browser page, then it will fire. But if they just click on "Home" or "Profile" or "Sign Out" to navigate to a different page within the React app, then the `beforeunload` event **will not** occur. The more React-y way to handle this is to call this `handleBeforeUnload` function (or just write those lines in-line) in the `return` value of the `useEffect` so that they are called as clean up steps when the component is being unmounted. This will also work on the `Album.js` modal, so as soon as someone closes the modal, that will emit a "user has left the room" event.

src/App.js
NatachaKey: @akosasante thank you, Akos, I just left the first approach- that grabs _albumId_ from useParams(), so I deleted the prop that Album.js passed before to ListeningAlbum,.js

src/components/album/ListeningAlbum.js
NatachaKey: @akosasante sure!!

NatachaKey: @akosasante Akos, I just noticed one thing-could you please take a look- I think in AlbumChatBox.js and in my code we are using different references to grab the user's id- i use 
  const { user } = useAuth();
const userId = user?._id;
and in albumChatBox.js -   const loggedInUserId = user?.id;
does it work in the same way- no matter what we use **_id** or just **id**?
https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/ef4ddbe8907bc63366ab62125db37e5524cac3b0/src/components/album/AlbumChatBox.js#L15C2-L15C35

src/components/album/ListeningAlbum.js
NatachaKey: @akosasante i think this is it, //still trying to figure our how it works, will now study the videos/refs you've left in this ticket https://codethedreamteam4-2023.atlassian.net/browse/PS-187

src/components/album/ListeningAlbum.js
```diff
@@ -0,0 +1,72 @@
...
+      const userId = user?._id;
+      socket.emit('listening-to-album-pause', { albumId, userId });
+    }
+  };
+
+  // Function to handle page unload
+  const handleBeforeUnload = () => {
+    if (socket) {
+      console.log('user left page');
+      socket.emit('user-left-page');
```
akosasante, line 36: Are we acting on this socket message anywhere or planning to? I guess to remove the user from the socket room maybe?

src/components/album/ListeningAlbum.js
```diff
@@ -0,0 +1,72 @@
...
+        console.log('Connected to WebSocket server');
+      });
+      setSocket(newSocket);
+    }
+
+    // disconnect from the socket (clean up step  when the component unmounts)
+    return () => {
+      handleBeforeUnload();
+    };
+  }, [albumId, socket, user?._id]);
```
akosasante, line 55: Linter tells us we should also add `handleBeforeUnload` as a dependency of this useEffect hook.

But since right now that is a plain function, it will cause excess renders, because every time React re-renders this component, that function will be recreated. Instead we need to wrap `handleBeforeUnload` in a useCallback so that React knows its not changing unless its depedencies change:

```
  const handleBeforeUnload = useCallback(() => {
    if (socket) {
      console.log('user left page');
      socket.emit('user-left-page');
      socket.disconnect();
    }
  }, [socket]);
```

alternatively, we can move `handleBeforeUnload` as a plain function _inside_ of this `useEffect` block. Either way is fine.

src/components/album/ListeningAlbum.js
akosasante: Ah in the backend there is a socket event for `user-leaving-page` , is that what we meant here?

akosasante: Right now, the backend event is not taking a userId, but that will probably be helpful for us since then we can know on the frontend _which_ user left the listening room

akosasante (commented):
Looking good; some suggestions for next steps to build this out further:

backend:
- have an explicit `join:album_listening_room` event that just joins that socket to a given album id room, instead of doing it inside the play event. This way as soon as the page loads, we can signal to the backend that we want to receive events about this room, before the user even clicks anything
- update `user-leaving-page` event to get a userId from the incoming data, and return the userId+message in the emitted event
- extra optional: return the profile image url of the user who pressed play

frontend
- add a `listeningUsers` useState hook to track and display which users are listening to the music
- inside the `newSocket.on('connect')` block add listeners for the `newSocket.on('play', ...)`, `newSocket.on('pause', ...)` and `newSocket.on('user-left', ...)` events that the backend emits. Then when some other user presses play, we can update the `listeningUsers` state variable and display their name or image on the page (you can refer to `AlbumChatBox.js` to see how something similar is done there).

PR #80, opened 2024-02-07
main page changed to home page

NatachaKey:


src/App.js
```diff
@@ -60,7 +60,7 @@ function App() {
       <Routes>
         <Route
           path="/"
-          element={isLoggedIn ? <Home /> : <Navigate to="/signIn" />}
+          element={isLoggedIn ? <Home /> : <Navigate to="/home" />}
         />
```
akosasante: Works well; but we can shorten this even more. If we're going to render the Home page either way whether or not they are logged in, then it doesn't make sense to even do the check or to do a full page navigation. Instead we can simplify to:

```suggestion
          <Route path="/" element={<Home />} />
```

_or_ we can even use react-routers syntax for "optional segments" where we follow it by a question mark: https://reactrouter.com/en/main/route/route#:~:text=You%20can%20have%20optional%20static%20segments%2C%20too%3A

```suggestion
          <Route path="/home?" element={<Home />} />
```

^ with this we could remove line 65, since the above suggestion would match both for `"/"` _and_ for `"/home"`

akosasante (approved):
Works well but I've left some suggestions for improvements.

akosasante:
We can now also remove the now-unused lines here (and any associated imports), as the linter will remind us: https://github.com/Code-the-Dream-School/dd-prac-team4-front/blob/homePageMoved/src/App.js#L22-L23

akosasante (changes requested):
Change request:

Now that the homepage is rendering the `AlbumList` component, that is causing `App.test.js` to fail because there are api calls being made that we're not mocking.

Let's update `mirageServer.js` to mock the `GET /albums/filter` and `POST /wishlists` calls that get made in the `AlbumList` component, so that our unit tests are not trying to make real HTTP tests within the test context. For now we can have both those mock endpoints return empty states (eg: for the list albums it can return: `{ albums: [] }`)

And to make our `App.test.js` more robust to actually verify that the component is rendering the `AlbumList` screen as we expect; we could add a couple lines that check for some text we expect to see. For example we can tell it to look for the input that has the label "Search Albums" on that page (using the async `findByLabelText` [function](https://testing-library.com/docs/queries/bylabeltext/)).

src/App.js
NatachaKey: @akosasante thank you for showing new react options, I will commit it now

NatachaKey:
@akosasante thank you, Akos! The requested changes have been performed, please take a look:)

akosasante (approved):
LGTM!  
  
Make sure you fix these lint warnings before merging though please:  
 

```
[9:53:44] ❯ npm run lint:check                

> bb-practicum-team4-front@0.1.0 lint:check
> eslint src

Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme

/Users/aasante/dev/practicum2023/frontend/src/App.js
   5:25  warning  'Navigate' is defined but never used. Allowed unused vars must match /^_/u             no-unused-vars
  23:9   warning  'isLoggedIn' is assigned a value but never used. Allowed unused vars must match /^_/u  no-unused-vars
```

NatachaKey:
@akosasante Thank you so much, Akos!!

PR #81, opened 2024-02-07
separated cart component leaving ordersidebar available

NatachaKey:


public/favicon.ico
akosasante, line 1: Nice, love the use of the favicon!

public/index.html
```diff
@@ -9,7 +9,8 @@
       name="description"
       content="Web site created using create-react-app"
     />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+    <link rel="apple-touch-icon" href="src\images\icons8-apple-music-48.png" />
```
akosasante: Everything in the `src` folder can only be accessed by other js/jsx files in the src folder (because they will be bundled together when we run `npm run build` ).

But this file `index.html` is served directly to the browser from the `public` folder. So it should refer to files that are either completely external, or also in the public folder. 

You can see this here: [https://realfavicongenerator.net/favicon_checker?protocol=https&site=beatbazaar.onrender.com%2F](https://realfavicongenerator.net/favicon_checker?protocol=https&site=beatbazaar.onrender.com%2F)

^ I deployed this commit on Render.com, and then ran this favicon checker. It is saying that it can't load the url [`https://beatbazaar.onrender.com/src\images\icons8-apple-music-48.png`](https://beatbazaar.onrender.com/src\images\icons8-apple-music-48.png)

(which makes sense since this is not a route that we have defined in our react router, so it just returns a 404 page).



We should instead copy the icon file into this `public` folder, and then update this path to be just the file name (the browser will automatically look in the public folder for it):

```suggestion
    <link rel="apple-touch-icon" href="%PUBLIC_URL%/icons8-apple-music-48.png" />
```

akosasante: We should still keep a copy of the file inside the src folder, since we need it when we refer to it from the Navbar component :+1:

public/index.html
```diff
@@ -9,7 +9,8 @@
...
+    <!-- <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" /> -->
```
akosasante: Is this commented-out line needed?

public/index.html
```diff
@@ -25,13 +26,21 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <!-- Material UI -->
-    <link
+    <!-- <link
```
akosasante: Can this commented-out line be removed?

src/components/Navbar.js
```diff
@@ -9,12 +9,15 @@ import {
...
 } from '@mui/material';
+import AddShoppingCartIcon from '@mui/icons-material/AddShoppingCart';
+import FavoriteIcon from '@mui/icons-material/Favorite';
 import { Link } from 'react-router-dom';
 import Logout from './userAuth/Logout';
 import { AuthStatus, useAuth } from '@akosasante/react-auth-context';
 import MenuIcon from '@mui/icons-material/Menu';
 import './Navbar.css';
-import Logo from '../images/Logo.png';
+//import Logo from '../images/Logo.png';
```
akosasante: This can be removed?

src/components/Navbar.js
```diff
@@ -31,17 +34,27 @@ const Navbar = ({ toggleDarkMode, mode }) => {
...
       }}
     >
       <Toolbar>
-        <img
-          src={Logo}
-          alt="Logo"
-          style={{ height: '2.5rem', marginRight: '0.625rem' }}
-        />
-        <Typography variant="h6" component="div" sx={{ flexGrow: 1 }}>
+        <Link component={Link} to="/home">
```
akosasante: `Link` does not have a property called `component`. [https://reactrouter.com/en/main/components/link#link](https://reactrouter.com/en/main/components/link#link)

Right now this is just adding an HTML attribute to the rendered `a` tag but other than that it's not doing anything. It can be removed.



```suggestion
        <Link to="/home">
```

src/components/Navbar.js
```diff
@@ -31,17 +34,27 @@ const Navbar = ({ toggleDarkMode, mode }) => {
...
+          <img
+            src={Logo}
+            alt="Logo"
+            style={{ height: '2.5rem', marginRight: '0.625rem' }}
+            component={Link}
```
akosasante: The html `img` tag does not have a valid attribute called `component`, this can be removed here too

src/components/Navbar.js
akosasante: Only MUI elements have a `component` property: [https://mui.com/material-ui/guides/composition/#component-prop](https://mui.com/material-ui/guides/composition/#component-prop)

It's usually used in Material-UI when you want to tell it to render a particular MUI component as a different html tag or component. For example in this file, on line 51-55, we are rendering a MUI component, `Typography` and telling React to render it as a `Link` MUI component (which makes it an `a` tag).

src/components/Navbar.js
```diff
@@ -9,12 +9,15 @@ import {
   Menu,
   MenuItem,
 } from '@mui/material';
+import AddShoppingCartIcon from '@mui/icons-material/AddShoppingCart';
```
akosasante: In my opinion, the AddCart icon looks like it's telling me I can click on it to _add more items_ to my cart. But really it takes me to a place to _view my cart items_. I wonder if a different MUI cart icon would be a better fit here? https://mui.com/material-ui/material-icons/?query=cart

src/components/Navbar.js
```diff
@@ -125,8 +144,8 @@ const Navbar = ({ toggleDarkMode, mode }) => {
                 <Button color="inherit" component={Link} to="/profile">
                   Profile
                 </Button>
-                <Button color="inherit" component={Link} to="/wishlist">
-                  Wishlist
+                <Button component={Link} to="/wishlist">
+                  <FavoriteIcon />
```
akosasante: Right now this is nested under the same `MenuItem` as the profile button above, so it looks like this:



That doesn't look quite right to me. If we want it to show up in its own row, we need to set it in its own menuitem in this array:

```suggestion
                <MenuItem key="wishlist" onClick={handleCloseUserMenu}>
                <Button component={Link} to="/wishlist">
                  <FavoriteIcon />
                </Button>
              </MenuItem>,
```

src/components/Navbar.js
```diff
@@ -144,3 +163,5 @@ Navbar.propTypes = {
   mode: PropTypes.oneOf(['light', 'dark']).isRequired,
 };
 export default Navbar;
+
+//NEED HELP -wishlist icon in navbar is not shown -
```
akosasante: Currently you added the wishlist icon under the `Menu` component which is only shown when you make the window small (ie: for mobile screens) as a hamburger menu on the right side of the navbar. 

If you wish to update the wishlist icon on the desktop navbar, you need to edit line 87

src/components/Navbar.js
```diff
@@ -58,9 +71,15 @@ const Navbar = ({ toggleDarkMode, mode }) => {
             </>
           )}
           <Button color="inherit" component={Link} to="/home">
-            Home
+            Shop
+          </Button>
+          <Button color="inherit" component={Link} to="/cart">
+            <AddShoppingCartIcon
```
akosasante: Do we want to show the shopping cart link even when a user is logged out? If not, we need to move this down below after line 84 where we check for logged in users.

src/components/Cart.js
```diff
@@ -0,0 +1,157 @@
...
+  const handleClearCart = () => {
+    dispatch(clearCart());
+  };
+
+  //64fd67815fca88489ab99564
+  const handleCheckout = () => {
+    // Create the order data object
+    const orderData = {
+      orderItems: Object.values(itemsInCart.items).map((item) => ({
+        album: item.album.id,
```
akosasante: Since basically all of this logic outside of the render (return value) is exactly the same as the logic in `src/components/OrderSidebar.js`, this seems like a good opportunity to extract this out so it can be reused. 

One way to approahc it would be to make each of these function calls a standalone function (eg: in a `utils/helper.js`) file, that can be called from both components. One difficulty we would run into though, is that these functions need to know about react/redux to work because they use `navigate`, `dispatch` , and `useSelector` functions. We _could_ pass in the needed values, like this:

```javascript
# utils/helpers.js
const hanldeMoveToWishlist = async (album, dispatch) => {
  ...
}

# Cart.js
import {handleMoveToWishlist} from 'src/utils/helpers.js'

const CartItems = () => {
  ...
  const dispatch = useDispatch();
  
  return {
    ...
    <Button onClick={() => handleMoveToWishlist(item.album, dispatch)}>
  }
}
```

but it's a little bit ugly. 

A better approach is if we want to make helper functions that also need to know about other hooks, is to make our own custom hooks: 

[https://www.freecodecamp.org/news/how-to-create-react-hooks/](https://www.freecodecamp.org/news/how-to-create-react-hooks/)

[https://react.dev/learn/reusing-logic-with-custom-hooks](https://react.dev/learn/reusing-logic-with-custom-hooks)

In this case, it could look something like:

```javascript
# src/redux/customHooks/useCart.js
export default function useCart() {
    const navigate = useNavigate();
    const dispatch = useDispatch();
    ...<copy all the shared code between Cart.js and OrderSidebar.js>...
    
    const variables = { itemsInCart, itemValues: Object.values(itemsInCart.items), subtotal, tax, total};
    const functions = { handleMoveToWishlist, handleAdd, handleReduce, handleRemove, handleClearCart, handleCheckout };

    return [variables, functions];
}

# in both Cart.js and OrderSidebar.js, the shared code can now be simplified to something like:
const CartItems() => {
  const [{itemValues, subtotal, tax, total}, { handleMoveToWishlist, handleAdd, handleReduce, handleRemove, handleClearCart, handleCheckout }] = useCart();
  
  return (...render code...)
}
```

src/components/Cart.js
```diff
@@ -0,0 +1,157 @@
...
+      total,
+    };
+    //clear the cart
+    dispatch(clearCart());
+    //navigate to checkout page and send order data as state
+    navigate('/checkout', { state: { orderData } });
+  };
+
+  return (
+    <div>
```
akosasante: I don't think we need to wrap this in a `div`. `Container` is already going to be a div anyway

src/components/album/AlbumsList.js
```diff
@@ -145,7 +145,7 @@ const AlbumsList = () => {
   };
 
   return (
-    <Container>
+    <Container fullWidth>
```
akosasante: `Container` component does not have a `fullWidth` property. Maybe you want to use `maxWidth` ? [https://mui.com/material-ui/api/container/#container-prop-maxWidth](https://mui.com/material-ui/api/container/#container-prop-maxWidth)

Currently this is causing this error in the console:

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
```diff
@@ -80,7 +80,7 @@ const AddToWishlistButton = ({ album, wishListId }) => {
     <>
       <Tooltip title={isAdded ? 'Remove from wishlist' : 'Add to wishlist'}>
         <IconButton
-          color="secondary"
+          color="error"
```
akosasante: This is fine; but a bit misleading to call this `error`. And if we ever want to change the color of `"error"` in our theme for the app, then it might be a bit of a surprise that it also changes the wishlist icon color. We can avoid this by creating a new color variable for `wishlist` in our MUI light and dark themes in `App.js`:

```javascript
const lightTheme = createTheme({
    palette: {
      mode: 'light',
      primary: { main: '#000000' },
      background: { default: '#C0C0C2', paper: '#ECECEC' },
      // etc any other customizations that we want to make for light mode
      wishlist: {
        main: '#ff4536',
      },
    },
    typography: {
      fontFamily: 'Poppins, sans-serif',
    },
  });
```

and then here we can refer to it with:

```suggestion
          color="wishlist"
```

src/components/user/UserProfile.js
```diff
@@ -47,7 +47,7 @@ export default function PersonalProfile() {
       <Link to="/updateUserInfo">
         {' '}
         {/* Specify the path where you want to redirect */}
-        <Button color="warning" style={{ backgroundColor: 'lightgrey' }}>
+        <Button color="success" style={{ backgroundColor: 'lightgrey' }}>
```
akosasante: this color is a bit difficult to read in dark mode. Maybe we just use the default colour set here? Or use a different background colour. Or create some custom colors in the theme if you refer: https://stackoverflow.com/questions/46486565/material-ui-customize-button-color

akosasante (commented):
Looking really great Natalia!

public/favicon.ico
NatachaKey: @akosasante thank you so much, Akos!! I am so glad you like it 🥰

public/index.html
NatachaKey: @akosasante thanks for the explanation, Akos! I didn't realize the importance of the location from which a file is served, see the difference now

public/index.html
NatachaKey: not at all, deleted👌

NatachaKey: @akosasante just a small doubt about the difference between these two tags, since the first one is a common favicon and the second one is for iOS devices- so the first should have an .ico extention to be served correctly in most browsers and the second  should be  .png , because iOS supports more this format, right? 
image

public/index.html
NatachaKey: done🙌

src/components/Navbar.js
NatachaKey: I see.. what I put here didn´t make any sence🫣

src/components/Navbar.js
NatachaKey: @akosasante totally agree, I think the new one looks better!

src/components/Navbar.js
NatachaKey: done!

src/components/Navbar.js
```diff
@@ -58,17 +71,23 @@ const Navbar = ({ toggleDarkMode, mode }) => {
             </>
           )}
           <Button color="inherit" component={Link} to="/home">
-            Home
+            <MusicVideoIcon/>
+          </Button>
```
akosasante: Looks great! 

One thing we should do for accessibility and making it clear to all users what this icon does, is add an `aria-label` (this will be read to screenreaders, and will show if the icon fails to load) and wrap this in a [Tooltip](https://mui.com/material-ui/react-tooltip/) component so that a tooltip shows when you hover over the button. We should do this for all our navbar buttons that are just icons now with no text.

```suggestion
          <Tooltip title="Home" arrow placement="bottom">
            <IconButton color="inherit" component={Link} to="/home" aria-label="link to home page">
              <MusicVideoIcon />
            </IconButton>
          </Tooltip>
```

akosasante: In my suggestion above I also switched the use of the `Button` component for an `IconButton` component, but either is really fine here.

src/components/Navbar.js
akosasante: Looks good! 

Do we intentionally want only the wishlist to be an icon button in the hamburger menu button for mobile? I think it would be better to keep it consistent, either have the whole hamburger menu be words or whole hamburger menu be icons. But I'm not super-design-minded, so I'm happy to proceed however you prefer!

src/components/Navbar.js
```diff
@@ -144,3 +165,5 @@ Navbar.propTypes = {
   mode: PropTypes.oneOf(['light', 'dark']).isRequired,
 };
 export default Navbar;
+
+//NEED HELP -wishlist icon in navbar is not shown -
```
akosasante: Comment can be removed now this is resolved :+1:

src/components/Navbar.js
akosasante: ^ I'm still curious if we intend to show the cart link for logged out users @NatachaKey

src/components/Navbar.js
NatachaKey: @akosasante thanks for the advice, Akos!! I totally agree with you that it´s better to stick to the same pattern, I am a desastrous designer haha so I appreciate any your idea or suggestion!❤️

src/components/Navbar.js
NatachaKey: resolved

src/components/Navbar.js
NatachaKey: @akosasante I was unsure about this.. It might be more comfortable to leave it as is if we had a usual online store with products, clothes, etc., so the user could add items to the cart without logging in, and log in only before checkout and payment. but since we have a music store and the user is already able to listen to music without logging in, there is no need in having this cart button for everyone. I’ll move it in the next commit🤲

src/components/Cart.js
NatachaKey: @akosasante thank you so much for the explanation and the article, Akos, I learned new info and now refactored the code, so from now on we can reuse this custom hook useCart in other parts of code when necessary, I will commit now

src/components/album/AlbumsList.js
NatachaKey: @akosasante thank you, Akos! I was trying to remove extra spaces from the right and left side of the page _on extra large screens_  (what we can see on the image below) and I´ve been trying many things, but **maxWidth** prop you suggested, worked out! thank you! 
Please tell me if it´s not looking good now because I´m not sure..
image

src/components/layout/AddToWishlistButton/AddToWishlistButton.js
NatachaKey: thank you! done!

src/components/user/UserProfile.js
NatachaKey: @akosasante, I totally agree, these styles need to be changed, I noted it to create a separate PR, at the moment I will leave the default color

src/components/Navbar.js
NatachaKey: done!

NatachaKey:
@akosasante Akos, please take a look when you have a moment, I've added the changes, and will move on to the next PR🙌

PR #82, opened 2024-02-09
purchased albums avalable in users profile

NatachaKey:


src/components/review/ReviewsList.js
akosasante, line 1: I guess a bit unrelated to this PR but since we have some changes here -- it would be nice to reset the `updateClicked` state to false after the albums are refreshed so that it's more clear that the edit was successful:

savechanges

src/components/user/UserProfile.js
```diff
@@ -9,13 +9,48 @@ import {
...
+      const response = await axiosInstance.get(`/users/showMe/withAlbums`);
+      console.log('this is what we get', response.data.user.purchasedAlbums);
+      setPurchasedAlbums(response.data.user.purchasedAlbums);
+      setLoading(false);
+    } catch {
+      console.error('Error fetching purchased albums:', error);
+      setError('Error fetching albums for this user. Please try again later.');
+      setLoading(false);
+    }
+  }, [user.userId]);
```
akosasante, line 33: Since we use the `error` state variable, this callback hook needs to know when it changes, so our linter is warning us that it's missing from the dependency array. 

At the same time, we don't use `user.userId` inside the useCallback hook so our linter will tell us that this does not need to be in the dependency array at all actually.:

```suggestion
  }, [error]);
```

src/components/user/UserProfile.js
```diff
@@ -9,13 +9,48 @@ import {
...
+  //       const response = await fetch('/api/showMe/withAlbums');
+  //       const data = await response.json();
+  //       setPurchasedAlbums(data.user.purchasedAlbums);
+  //     } catch (error) {
+  //       console.error('Error fetching purchased albums:', error);
+  //     }
+  //   };
+
+  //   fetchPurchasedAlbums();
+  // }, []);
```
akosasante, line 52: We could write it this way with all the callback code in a useEffect instead of a separate hook. That would be totally fine as well.

The main things to look out for with what you've commented out is making sure we've set the error and loading state variables, and we'd want to use `axiosInstance` since that one already has our base url set up to work in both local dev and in the deployed code, and is already passing along the authentication cookie. We could do the same with `fetch` but it's unnecessary extra work at this point in the application.

src/components/user/UserProfile.js
```diff
@@ -79,6 +114,54 @@ export default function PersonalProfile() {
...
+      <Table className="text-center mt-5">
+        <tbody>
+          <TableRow>
+            <TableCell>
+              {' '}
+              <b>
+                {' '}
+                <em></em>
+              </b>
+            </TableCell>
```
akosasante, line 131: We could just have an empty `TableCell` component here to represent the empty header for this column:
```suggestion
            <TableCell/>
```

src/components/user/UserProfile.js
```diff
@@ -79,6 +114,54 @@ export default function PersonalProfile() {
...
+              <div>
+                <TableCell>
+                  <img
+                    src={purchasedAlbum.album.image}
+                    alt="albumImage"
+                    height="100px"
+                    width="100px"
+                  />
+                </TableCell>
+              </div>
```
akosasante, line 157: I don't think we need this extra `div` wrapping the table cell -- it also kind of breaks the border line between the rows right now:

```suggestion
              <TableCell>
                <img
                  src={purchasedAlbum.album.image}
                  alt="albumImage"
                  height="100px"
                  width="100px"
                />
              </TableCell>
```

akosasante (commented):
Looking great, just left a few notes!

